PR #2, opened 2024-08-27
Progress on image tile decoding

kylebarron:
Download the test file to the root of the repo:

```
aws s3 cp s3://naip-visualization/ny/2022/60cm/rgb/40073/m_4007307_sw_18_060_20220803.tif ./ --request-payer
```

Then run the tests (`cargo test`)

Then run this Python code:

```py
import numpy as np
from rasterio.plot import show, reshape_as_raster

with open("img.buf", "rb") as f:
    buffer = f.read()

arr = np.frombuffer(buffer, np.uint8)
# We first need to reshape into "image" axes, then we need to reshape as "raster" axes.
arr = arr.reshape(512, 512, 3)
arr = reshape_as_raster(arr)
show(arr, adjust=True)
```
image


src/decoder.rs
```diff
@@ -0,0 +1,133 @@
...
+            let mut reader = reader;
+            reader.read_exact(&mut [0; 2])?;
+
+            Box::new(Cursor::new(&jpeg_tables[..jpeg_tables.len() - 2]).chain(reader))
+                as Box<dyn Read>
+        }
+        None => Box::new(reader),
+    };
+
+    let mut decoder = jpeg::Decoder::new(jpeg_reader);
```
geospatial-jeff, line 58: this is so much better than python 😅

src/ifd.rs
```diff
@@ -504,12 +533,24 @@ impl ImageFileDirectory {
         }
     }
 
-    pub async fn get_tile(&self, x: usize, y: usize) {
+    pub async fn get_tile(
```
geospatial-jeff, line 536: Small NIT but I'd recommend breaking this up into two separate methods, something like `get_tile` and `get_and_decode_tile`.  The first would parse the header and send a range request for the desired offset, the second is a wrapper around the first and `decode_tile`.

This does a better job decoupling IO/CPU bound code by allowing the caller to easily perform the range request without also decompressing the data.

kylebarron: Yes, that makes a lot of sense. In general, optimizing IO/CPU handling is something I want to leave for future PRs after we have the general prototype working and some tests passing.

PR #5, opened 2025-02-20
Define async reader interface

kylebarron:
Closes https://github.com/developmentseed/aiocogeo-rs/issues/4

### DIFF
--- a/src/async_reader.rs
+++ b/src/async_reader.rs
@@ -1,0 +1,113 @@
+use bytes::Bytes;
+use futures::future::{BoxFuture, FutureExt, TryFutureExt};
+use object_store::ObjectStore;
+use std::io::SeekFrom;
+use std::ops::Range;
+use std::sync::Arc;
+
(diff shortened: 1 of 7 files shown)


src/async_reader.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct ObjectReader {
+    store: Arc<dyn ObjectStore>,
+    path: object_store::path::Path,
+}
+
+impl ObjectReader {
+    /// Creates a new [`ParquetObjectReader`] for the provided [`ObjectStore`] and [`ObjectMeta`]
```
vincentsarago, line 85: parquet?

kylebarron: Good catch (this was ported from the parquet crate)

kylebarron:
```suggestion
    /// Creates a new [`ObjectReader`] for the provided [`ObjectStore`] and path
```

PR #18, opened 2025-02-25
Prefetch reader

kylebarron:
Allows for fetching first X kb of a file, and then delegating further requests to this cached buffer.

This caused the Python file open to reduce from 2.5 minutes to 0.7sec

cc @geospatial-jeff. Because this is implemented as a middleware, we could implement whatever sort of cache we want here, like a block cache.

### DIFF
--- a/src/async_reader.rs
+++ b/src/async_reader.rs
@@ -114,6 +114,44 @@ impl AsyncFileReader for ObjectReader {
     }
 }
 
+pub struct PrefetchReader {
+    reader: Box<dyn AsyncFileReader>,
+    buffer: Bytes,
+}
+
+impl PrefetchReader {
+    pub async fn new(mut reader: Box<dyn AsyncFileReader>, prefetch: u64) -> Result<Self> {
+        let buffer = reader.get_bytes(0..prefetch).await?;
+        Ok(Self { reader, buffer })
+    }
+}
+
+impl AsyncFileReader for PrefetchReader {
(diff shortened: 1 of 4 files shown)


src/async_reader.rs
```diff
@@ -114,6 +114,44 @@ impl AsyncFileReader for ObjectReader {
...
 }
 
+pub struct PrefetchReader {
+    reader: Box<dyn AsyncFileReader>,
+    buffer: Bytes,
+}
+
+impl PrefetchReader {
+    pub async fn new(mut reader: Box<dyn AsyncFileReader>, prefetch: u64) -> Result<Self> {
+        let buffer = reader.get_bytes(0..prefetch).await?;
```
kylebarron, line 124: We pre-fetch this prefix of the file when "opening" this reader layer

src/async_reader.rs
```diff
@@ -114,6 +114,44 @@ impl AsyncFileReader for ObjectReader {
...
+    }
+}
+
+impl AsyncFileReader for PrefetchReader {
+    fn get_bytes(&mut self, range: Range<u64>) -> BoxFuture<'_, Result<Bytes>> {
+        if range.start < self.buffer.len() as _ {
+            if range.end < self.buffer.len() as _ {
+                let usize_range = range.start as usize..range.end as usize;
+                let result = self.buffer.slice(usize_range);
+                async { Ok(result) }.boxed()
```
kylebarron, line 135: Then if the start and end of the requested range are fully within the existing cached buffer, we return that slice.

PR #52, opened 2025-03-14
Remove print and debug statements

maxrjones:


kylebarron:
Thanks; 

Looks like they slipped in in https://github.com/developmentseed/async-tiff/pull/30.

I think there's some clippy lint to assert that there are no `dbg` statements; but not sure how to turn that on

PR #61, opened 2025-03-17
Fix wheels; fetch submodule

kylebarron:
Trying to fix the error of 
```
💥 maturin failed
  Caused by: Failed to add the files to the wheel
  Caused by: Failed to add the python module to the package
  Caused by: No such file or directory (os error 2)
```
it's a confusing error but then I got a hunch that it's probably because the submodule isn't fetched, so the symlink can't be followed.


kylebarron:
Testing here: https://github.com/developmentseed/async-tiff/actions/runs/13906148851

PR #71, opened 2025-03-20
Feature gate `object_store` and `reqwest`

feefladder:
As discussed in #6, only feature gate and not pulling coalesce_ranges in-tree


src/error.rs
```diff
@@ -32,6 +33,7 @@ pub enum AsyncTiffError {
     InternalTIFFError(#[from] crate::tiff::TiffError),
 
     /// Reqwest error
+    #[cfg(feature = "reqwest")] // see https://www.reddit.com/r/rust/comments/xyik51/comment/irhei39/
```
kylebarron, line 36:
```suggestion
    #[cfg(feature = "reqwest")]
```

Cargo.toml
```diff
@@ -15,14 +15,19 @@ flate2 = "1.0.20"
...
 weezl = "0.1.0"
 
 [dev-dependencies]
 tiff = "0.9.1"
 tokio = { version = "1.9", features = ["macros", "fs", "rt-multi-thread"] }
+
+[features]
+default = ["object_store"]
+reqwest = ["dep:reqwest"]
+object_store = ["dep:object_store", "reqwest"]
```
kylebarron, line 33:
```suggestion
default = ["object_store", "reqwest"]
reqwest = ["dep:reqwest"]
object_store = ["dep:object_store"]
```
In the future `object_store` may not depend on `reqwest`

kylebarron (commented):
Can you add to CI variants that test that `async-tiff` compiles with each combination of features? 

You can see how parquet-wasm uses cargo-all-features
 https://github.com/kylebarron/parquet-wasm/blob/e85322b592ab224938861895a06f4f3947a77802/Cargo.toml#L117

https://github.com/kylebarron/parquet-wasm/blob/e85322b592ab224938861895a06f4f3947a77802/.github/workflows/test.yml#L38-L41

feefladder:
just wanted to not: removing `object_store` from default deps makes tests fail, which is fine

feefladder:
it was not fine and is fixed now

src/reader.rs
```diff
@@ -67,6 +70,21 @@ impl AsyncFileReader for Box<dyn AsyncFileReader + '_> {
...
+            file.seek(std::io::SeekFrom::Start(range.start))?;
+            let len = (range.end - range.start) as usize;
+            let mut buf = vec![0u8; len];
+            file.read_exact(&mut buf)?;
+            let res = Bytes::copy_from_slice(&buf);
+            Ok(res)
+        }
+        .boxed()
+    }
+}
```
kylebarron, line 86: We should not add this because this will perform blocking reads on the main thread. 

Instead we should probably restore the tokio integration 
https://github.com/developmentseed/async-tiff/blob/6a7a2dc038f7b84e8ada5ef16e2cb8c1e5504334/src/reader.rs#L70-L91 which will allow reading from a [tokio async file](https://docs.rs/tokio/latest/tokio/fs/struct.File.html), which I think performs file IO from a thread pool

Cargo.toml
```diff
@@ -31,3 +31,11 @@ tokio = { version = "1.9", features = ["macros", "fs", "rt-multi-thread"] }
 default = ["object_store", "reqwest"]
 reqwest = ["dep:reqwest"]
 object_store = ["dep:object_store"]
+
+[package.metadata.cargo-all-features]
```
kylebarron, line 35: You can't only add the metadata, you need to actually add it as a step to CI

src/reader.rs
feefladder: The problem I had was that the feature testing also wants to test with all features disabled, and I didn't exactly know how to deal with that... added a fallback in util.rs

feefladder: Also tokio's asyncread require `&mut self`, so I changed the impl to `tokio::fs::File`, where we can `try_clone()` on `&self`

feefladder:
@kylebarron should be done!

Cargo.toml
```diff
@@ -15,14 +15,27 @@ flate2 = "1.0.20"
...
 num_enum = "0.7.3"
-object_store = "0.12"
+object_store = { version = "0.12", optional = true }
 # In the future we could make this feature-flagged, but for now we depend on
 # object_store which uses reqwest.
-reqwest = { version = "0.12", default-features = false }
+reqwest = { version = "0.12", default-features = false, optional = true }
 thiserror = "1"
-tokio = { version = "1.43.0", optional = true }
+tokio = { version = "1.43.0", optional = true, features = ["fs", "io-util"] }
```
kylebarron, line 23: Do we use these features? I think these should be enabled only on the dev-dependency `tokio`

tests/image_tiff/util.rs
```diff
@@ -1,15 +1,78 @@
...
+pub(crate) async fn open_tiff(filename: &str) -> TIFF {
+    // let store = Arc::new(LocalFileSystem::new_with_prefix(current_dir().unwrap()).unwrap());
+    let path = format!("{TEST_IMAGE_DIR}/{filename}");
+    let reader = Arc::new(
+        tokio::fs::File::open(path)
+            .await
+            .expect("could not open file"),
+    );
+    TIFF::try_open(reader).await.unwrap()
+}
```
kylebarron, line 43: We shouldn't add this test, because all this is doing is testing that the same file can be read with tokio. But as long as the traits compile we can use only `object_store` at runtime.

We should always use `object_store` for testing, and so `object_store` should always be included in `dev-dependencies`.

PR #72, opened 2025-03-21
Wasm futures

feefladder:
depends on #71 
Make the futures of the trait wasm-compatible

This uses `async_trait` crate, but the function signatures remain largely untouched...
change is from
```rust
fn get_bytes(&self, range: Range<u64>) -> BoxFuture<'_, AsyncTiffResult<Bytes>>
// which afaik equals
fn get_bytes<'a>(&'a self, range: Range<u64>) -> BoxFuture<'a, AsyncTiffResult<Bytes>>
```
to
```
fn get_bytes<'life0, 'async_trait>(&'life0 self, range: Range<u64>) -> BoxFuture<'async_trait, AsyncTiffResult<Bytes>>
where:
  'life0: 'async_trait,
  Self: 'async_trait
```
So I think it's not that big of a change


kylebarron:
Can you explain how the `BoxFuture` is not already wasm-compatible? The [`ObjectStore` trait](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html) uses `BoxFuture`, and it is wasm32 compatible.

feefladder:
So there's wasm and wasm... In browser, where everything becomes a Javascript object, [futures are ?Send](https://users.rust-lang.org/t/is-it-possible-to-take-a-jsfuture-as-send/97774/2)

kylebarron:
This `AsyncFileReader` trait is identical to the `parquet` crate's [`AsyncFileReader`](https://docs.rs/parquet/latest/parquet/arrow/async_reader/trait.AsyncFileReader.html), which I am using in production in wasm32-unknown-unknown in the browser: https://github.com/kylebarron/parquet-wasm

feefladder:
interesting...

feefladder:
Ok, [you're right](https://github.com/kylebarron/parquet-wasm/pull/96#issuecomment-1107925910), I'm wrong...
seems wasm compatibility could be achieved with another middleware:
```
pub struct WasmReader {
  inner: Arc<dyn NotSendAsyncFileReader>
}

impl AsyncFileReader for WasmReader {
  async fn get_bytes(&self, range: Range<u64>) -> BoxFuture<'_, AsyncTiffResult<Bytes>> {
    let (rx, tx) = oneshot::new()
    let inner = self.inner.clone()
    async move {
      tx.send(inner.get_bytes(range).await) // this task is not Send, but that's ok since we don't use any references
    }
    rx.await // channel receive is Send!!!
  }
}
```

There's quite some weeds also, where wasm-bindgen plays a lot of magic tricks, but in the end it was indeed parquet-wasm that had the solution :D

anyways.... ~~Would it maybe be nice to use the [1.75 announced trait_variant](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#async-fn-in-public-traits)? That way it's easy~~ here is dynamic dispatch...

kylebarron:
It's not really a middleware because it's not wrapping anything. You just need this implementation: https://github.com/kylebarron/parquet-wasm/blob/e85322b592ab224938861895a06f4f3947a77802/src/reader_async.rs#L230-L291

kylebarron:
I think we established that we don't need the changes in this PR for wasm support, so I'll close this

feefladder:
> It's not really a middleware because it's not wrapping anything. You just need this implementation: https://github.com/kylebarron/parquet-wasm/blob/e85322b592ab224938861895a06f4f3947a77802/src/reader_async.rs#L230-L291

The part that makes it wasm-compatible (turns a non-`Send` js future into a `Send`able Rust future is in the [`make_range_request_with_client` implementation](https://github.com/kylebarron/parquet-wasm/blob/e85322b592ab224938861895a06f4f3947a77802/src/reader_async.rs#L436-L447) (l436-L447), and it does the channel-trick. In my thoughts this _could_ be wrapping a default `make_request` or something, but it is indeed up to the user to implement.

PR #76, opened 2025-03-23
Read tag data in its entirety from AsyncFileReader.

feefladder:
Read tags from a buffer which is fetched in its entirety in stead of reading them value-by-value. This allows any middleware to be a bit smart about caching outside of the prefetch range. Also it simplifies logic in `read_tag_value`.

This same idea could also be used for pre-fetchin/reading IFDs themselves from the point the count is known. That _does_ change things a tiny little more, so is its own PR

### DIFF
--- a/src/ifd.rs
+++ b/src/ifd.rs
@@ -839,8 +839,6 @@ impl ImageFileDirectory {
 
 /// Read a single tag from the cursor
 async fn read_tag(cursor: &mut AsyncCursor, bigtiff: bool) -> AsyncTiffResult<(Tag, Value)> {
-    // let start_cursor_position = cursor.position();
-
     let tag_name = Tag::from_u16_exhaustive(cursor.read_u16().await?);
 
     let tag_type_code = cursor.read_u16().await?;
@@ -854,10 +852,6 @@ async fn read_tag(cursor: &mut AsyncCursor, bigtiff: bool) -> AsyncTiffResult<(T
     };
 
     let tag_value = read_tag_value(cursor, tag_type, count, bigtiff).await?;
-
-    // TODO: better handle management of cursor state <- should be done now
-    // let ifd_entry_size = if bigtiff { 20 } else { 12 };
-    // cursor.seek(start_cursor_position + ifd_entry_size);
 
     Ok((tag_name, tag_value))
 }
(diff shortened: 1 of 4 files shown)


kylebarron:
Can you summarize what you're trying to do here? The changes aren't clear.

> Read tags from a buffer which is fetched in its entirety

How do you know the full byte range of the IFD?

feefladder:
(depends on #71)

So I cooked up a little example that read a cog over http and it was reading all tag data one after the other, requesting it from the underlying reader one at a time. We know a lot more than the underlying reader, so we can "optimize" for that.

- From the moment we know the tag type and count, we can know the _tag data_ byte size `tag_value_byte_size = tag_type.size() * count`
  - If that fits in the offset:
    - create a reader over part of the offset
    - advance the cursor to the end of offset
  - Otherwise:
    - read the offset, putting the cursor in the right position
    - fetch the data directly using the underlying `AsyncFileReader` without touching the cursor.
- now we are working with a decoupled `EndianAwareReader` that holds (only) all tag data.
- All later logic only needs to deal with count 1 or more, since it will either be:
  - count 1: `Value::Whatevs`
  - count more: `Value::List(Vec<Value::Whatevs>)`
The later logic is partially re-used from earlier, so the diff is a bit of a mess, but afterwards, there are only two match statements, for count 1 or other.

So if we are reading out-of-prefetch or without any middleware, it will at most make 1 request per IFD data, in stead of `count` requests.

In `tiff2`, I implemented this so we could re-use the logic for sync and async, but then it turned out to also be a good speedup/reduce the number of requests.

The IFD size can also be known beforehand, when `bigtiff` and `tag_count` are known, reducing the unoptimized number of requests for ifds as well.

src/ifd.rs
```diff
@@ -878,56 +878,45 @@ async fn read_tag_value(
...
+        let offset = if bigtiff {
+            cursor.read_u64().await?
+        } else {
+            cursor.read_u32().await?.into()
+        };
+        let reader = cursor.reader().get_bytes(offset..offset+value_byte_length).await?.reader();
+        EndianAwareReader::new(reader, cursor.endianness())
+        // cursor.seek(offset);
+        // cursor.read(value_byte_length).await?
+    };
```
feefladder, line 906: this is the main logic, after which there are only 2 match statements

kylebarron:
> AsyncCursor, which reads single values[1](#user-content-fn-1-cadcf7c120af874b977641c2820602e4)

Yes it reads single values, but that's **by design**. It makes the underlying code simple to reason about, while deferring all of the actual request logic to the network layer. 

The `PrefetchReader` is my answer to this. We could extend the prefetch reader to fetch sequences of X kb, so if you create a `PrefetchReader` for the first 16kb, but then you request a byte range 20,000-21,000, it would fetch the second 16kb, or something like that.

I think it's possibly/likely a good idea to tell the network layer when you're doing image vs metadata requests, but I'm not sure about some of these other more invasive code changes.

> So if we are reading out-of-prefetch or without any middleware, it will at most make 1 request per IFD data, in stead of `count` requests.

Right but my point is that at most one of these async functions should ever request data anyways. 

I'd be happy to make the docs more explicit to say that the underlying reader should never be used without a caching/buffering middleware.

> This PR is an improvement, since it fetches the entire array in one go

The overhead of an async function is likely nanoseconds, so changing the number of async calls (as long as the underlying request layer responds from its cache) should be insignificant.

kylebarron:
Ok first things first you need to designate **one** PR to review, instead of four, and ensure that is up to date. I can't even tell how many things are changing in this PR.

feefladder:
Ok, will rebase on top of #79, is that ok?

PR #77, opened 2025-03-23
Also buffer IFD reads

feefladder:
Builds on #76, but also computes ifd size beforehand and fetches all of it. Only really beneficial for non-cogs.

The same:
- calculate IFD size = `ifd_entry_byte_count * tag_count`
- load into `EndianAwareReader`
- read the rest of the IFD using the `EndianAwareReader`


kylebarron:
If this builds on #76 we can come back to discuss this after we resolve #76 and #78

feefladder:
Also, this was just a thought I had. We can also close it, because it's mostly only useful for non-cog tiffs that have their IFDs not in the prefetch range, which could well be solved by a simple buffering reader. (I sense some resistance to #76, but that one is more important)

PR #81, opened 2025-03-27
Read tag data in its entirety from AsyncFileReader. 

### LINKED ISSUE #70: Handling large IFD data arrays
EDIT: Since deferring fetching of `tile_info=tile_offsets+tile_byte_counts` was quite an intrusive change, and the remaining `tile_info`s can be reasonably estimated from the first `tile_offsets`, this issue changed a bit in my question and is now mainly about #76 

TileByteCounts and TileOffsets arrays are special-cased in the [COG spec](https://docs.ogc.org/is/21-026/21-026.html#_requirement_reduced_resolution_subfiles), and don't generally fit within the first 16kB.

Thus, using `coalesce_ranges` and waiting with fetching that data until all tags are read could be beneficial (reduce the number of requests). I've updated the amazing diagram to include COG file structure:

Image

feefladder:
Why was #76 closed 👀? 

Closes #70 

This PR allows a caching middleware to directly estimate the remaining `TileOffsets` and `TileByteCounts` arrays from the byte size of `TileOffsets0`, which is in #74 as a proof-of-concept.

- without this: first out-of-cache request will be a single u64 byte range
- with this: first out-of-cache request will be the full `TileOffsets0` array

~~A middleware _could_ do exponential cache growing, just like a `Vec`, but if a good estimate (`2*range_len+4*range_len.isqrt()`) of size can be made, `Vec::with_capacity(good_estimate)` is better [derivation.pdf](https://github.com/user-attachments/files/19487170/main.pdf).~~

**EDIT**: Actually, an exponential prefetch (exponent 2) will work quite well with this, since the byte size of all ifds ~>= `4*TileOffsets0.isqrt()`[^1]

also Cleans up tag loading logic and cursor handling.

[^1]: The estimate of $2b_{offset0}+4\sqrt(b_{offset0}$ is only based on the assumption that we are dealing with a square unmasked bigtiff cog. smalltiff adds a $\frac{4}{3}$ factor, masks a $2$ factor and square-ness only works on the $\sqrt{}$ factor, which is small in comparison.

### DIFF
--- a/src/ifd.rs
+++ b/src/ifd.rs
@@ -839,8 +839,6 @@ impl ImageFileDirectory {
 
 /// Read a single tag from the cursor
 async fn read_tag(cursor: &mut AsyncCursor, bigtiff: bool) -> AsyncTiffResult<(Tag, Value)> {
-    // let start_cursor_position = cursor.position();
-
     let tag_name = Tag::from_u16_exhaustive(cursor.read_u16().await?);
 
     let tag_type_code = cursor.read_u16().await?;
@@ -854,10 +852,6 @@ async fn read_tag(cursor: &mut AsyncCursor, bigtiff: bool) -> AsyncTiffResult<(T
     };
 
     let tag_value = read_tag_value(cursor, tag_type, count, bigtiff).await?;
-
-    // TODO: better handle management of cursor state <- should be done now
-    // let ifd_entry_size = if bigtiff { 20 } else { 12 };
-    // cursor.seek(start_cursor_position + ifd_entry_size);
 
     Ok((tag_name, tag_value))
 }
@@ -878,7 +872,17 @@ async fn read_tag_value(
         return Ok(Value::List(vec![]));
     }
 
-    let tag_size = tag_type.size();
+    let tag_size = match tag_type {
+        Type::BYTE | Type::SBYTE | Type::ASCII | Type::UNDEFINED => 1,
+        Type::SHORT | Type::SSHORT => 2,
+        Type::LONG | Type::SLONG | Type::FLOAT | Type::IFD => 4,
+        Type::LONG8
+        | Type::SLONG8
+        | Type::DOUBLE
+        | Type::RATIONAL
+        | Type::SRATIONAL
+        | Type::IFD8 => 8,
+    };
 
     let value_byte_length = count.checked_mul(tag_size).unwrap();
 
@@ -901,12 +905,10 @@ async fn read_tag_value(
         };
         let reader = cursor
             .reader()
-            .get_bytes(offset..offset + value_byte_length)
+            .get_metadata_bytes(offset..offset + value_byte_length)
             .await?
             .reader();
         EndianAwareReader::new(reader, cursor.endianness())
-        // cursor.seek(offset);
-        // cursor.read(value_byte_length).await?
     };
     // Case 2: there is one value.
     if count == 1 {
--- a/src/ifd.rs
+++ b/src/ifd.rs
@@ -7,7 +7,7 @@ use num_enum::TryFromPrimitive;
 
 use crate::error::{AsyncTiffError, AsyncTiffResult};
 use crate::geo::{GeoKeyDirectory, GeoKeyTag};
-use crate::reader::{AsyncCursor, AsyncFileReader};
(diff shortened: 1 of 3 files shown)


kylebarron:
> Why was #76 closed 👀?

#76 was auto-closed because when I wrote the description of #79 (https://github.com/developmentseed/async-tiff/pull/79#issue-2944634103) I had the words "I think this probably also closes https://github.com/developmentseed/async-tiff/pull/76"

kylebarron:
> directly estimate the remaining `TileOffsets` and `TileByteCounts` arrays from the byte size of `TileOffsets0`

This is a horribly leaky abstraction because you're trying to infer **how much more data you'll need to read in total** from **individual requests made to the underlying network layer**. 

On the contrary it's **the entire point** that these layers are isolated, so it's **intentional** that the network layer should not be able to infer where the header parsing is, solely based on requests. The internals of how the metadata is read **should be isolated** from reading the bytes.

It might work in this specific case to do your estimation, but that happens to rely on this specific case, and if the internals of how metadata is parsed is changed again in the future, that would break your code.

Instead of hacking around this via byte ranges, I think it would be better to expose a little more of the internals of metadata parsing, so that you can swap our your own implementation of reading the metadata if you wish. See #82

feefladder:
> > directly estimate the remaining `TileOffsets` and `TileByteCounts` arrays from the byte size of `TileOffsets0`
> 
> This is a horribly leaky abstraction because you're trying to infer **how much more data you'll need to read in total** from **individual requests made to the underlying network layer**.

based on the spec. That #83 seems to work in the wild is a good sign for #74 , since the only unspecified behaviour is the order of tile_info arrays and #83 assumes monotonic reads from the start of the file, which would only work if they are ordered like I assumed in my [derivation](https://github.com/user-attachments/files/19487170/main.pdf).

The use-case is a cog reader that can read (unmasked) small cogs' metadata with a single ~16kB prefetch and large cogs with at most 1 more fetch.

> On the contrary it's **the entire point** that these layers are isolated, so it's **intentional** that the network layer should not be able to infer where the header parsing is, solely based on requests. The internals of how the metadata is read **should be isolated** from reading the bytes.

#83 also assumes monotonic reads, otherwise the caching won't kick in since the requested range overlap. I think that's good, because it'll work for cogs and not mess up anything for non-cogs. It's just sad that with a 16kB prefetch it'll take 10 requests for a 4-8MB metadata COG in stead of 1.

feefladder:
Could we voice chat or something?

feefladder:
> On the contrary it's **the entire point** that these layers are isolated, so it's **intentional** that the network layer should not be able to infer where the header parsing is, solely based on requests. The internals of how the metadata is read **should be isolated** from reading the bytes.

Layers need to be isolated: Yes! Clearly, the exponential prefetch better separates layers.

Still, I think #83 works very well _together_ with this PR. (this PR would reduce additional requests from ~9 to 2).

> Instead of hacking around this via byte ranges, I think it would be better to expose a little more of the internals of metadata parsing, so that you can swap our your own implementation of reading the metadata if you wish. See #82

I completely errored on this and am sad I couldn't appreciate all the effort you put into #82 

> > If you think 4 match blocks that change cursor state to anywhere in the tiff is more maintainable than 2 match blocks that set the cursor at the end of the tag offset field, sure don't merge #81.
> 
> Yes, I think it's more maintainable because it's exactly the same as the [upstream code](https://github.com/image-rs/image-tiff/blob/36f7b3b4aa9ad3514017906bf4efdbf646df189f/src/decoder/ifd.rs#L375-L639). I don't want to have to think about the specifics of tag parsing. The code as it is **just works**, and even reviewing your changes is really hard because I would have to research the tag structure and validate that it's correct.

I've already made [unit tests](https://github.com/feefladder/tiff2/blob/b9c48bf8f9ec7843e73eacfb568cc139a8cec9e7/src/structs/entry.rs#L765-L1024) for tag parsing in tiff2. Would it help if:

1. I copy over the unit tests from tiff2 so they work on the current (upstream) implementation in a separate PR
2. this PR makes all unit tests still pass
3. show on a cog (link) you provide that it improves metadata fetching (3 or 4 requests) with a default 32kB prefetch and exponential prefetch.

PR #83, opened 2025-03-27
Exponential overflow header buffer

kylebarron:
### Changes 

- Updates `AsyncFileReader` to use `&mut self` instead of `&self` on the trait methods. This is necessary so that the buffer can be **mutated** when fetching the overflow, while also adding it to the cached buffers
- Use `Box` instead of `Arc` in reader code, now that we need mutability of the reader.
- Updates the tokio integration to implement `AsyncFileReader` directly on anything that implements `AsyncRead` and `AsyncSeek`. We don't need a wrapper object now that we use `&mut self` ourselves.
- Add test to verify overflow handling

This PR is not made on top of #82, but I think some mix of this and #82 makes sense.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -591,7 +591,7 @@ mod test {
         let reader = TestAsyncFileReader {
             buffer: Bytes::from_static(underlying_buffer),
         };
-        let mut prefetch_reader = PrefetchReader::new(Box::new(reader), 5, 1.1).await.unwrap();
+        let mut prefetch_reader = PrefetchReader::new(Box::new(reader), 5, 1.).await.unwrap();
 
         // Cached
         assert_eq!(
@@ -642,5 +642,10 @@ mod test {
                 .as_ref(),
             underlying_buffer
         );
+
+        // Assert underlying buffers were cached
+        assert_eq!(prefetch_reader.buffers[0].as_ref(), b"abcde");
+        assert_eq!(prefetch_reader.buffers[1].as_ref(), b"fghij");
+        assert_eq!(prefetch_reader.buffers[2].as_ref(), b"klmno");
     }
 }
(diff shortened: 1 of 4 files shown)


src/reader.rs
```diff
@@ -227,54 +207,121 @@ impl ReqwestReader {
...
+    async fn expand_prefetch(&mut self, range: Range<u64>) -> AsyncTiffResult<Bytes> {
+        let existing_buffer_length = self.buffer_length() as u64;
+        let additional_fetch =
+            (existing_buffer_length as f64).powf(self.overflow_fetch_exponent) as u64;
+
+        // Make sure that we fetch at least the entire desired range
+        let new_range =
+            existing_buffer_length..range.end.max(existing_buffer_length + additional_fetch);
+        let buffer = self.reader.get_metadata_bytes(new_range).await?;
+        self.buffers.push(buffer);
```
kylebarron, line 259: This is where we request the overflow and add it to the existing buffering (`self.buffers.push(buffer)`)

src/reader.rs
```diff
@@ -227,54 +207,121 @@ impl ReqwestReader {
...
+    fn buffer_length(&self) -> usize {
+        self.buffers.iter().fold(0, |acc, x| acc + x.len())
+    }
+
+    /// Access the buffer range out of the cached buffers
+    ///
+    /// ## Panics
+    ///
+    /// If the range does not fall completely within the pre-cached buffers.
+    fn buffer_slice(&self, range: Range<u64>) -> Bytes {
```
kylebarron, line 275: This reads from the existing buffers, handling a cross-buffer merge in case the input request overlaps multiple of `self.buffers`

kylebarron:
Closing in favor of #140

PR #85, opened 2025-03-31
added unit tests for tag reading

feefladder:
Add tests for reading tags.

Whenever a `Tag::Type`->`Value::Signed/Byte` conversion happens, this is indicated in a trailing comment.


kylebarron:
Looks good; can you fix the failing formatting?

PR #86, opened 2025-04-02
added predictors

feefladder:
Added predictors and rudimentary tests.

Since floating point predictors shuffle horizontal padding into the output, quite a lot more information is needed, so I made a public `PredictorInfo` struct with public methods that give tile/chunk info. 

Some notes:
- Also use a registry structure, so people can swap out their own if they'd want to
- I did put in the possibility for multiple sample formats, which links our lifetime to the tile... I think it'd also be perfectly fine to accept only a single value, but then I think it'd also make sense to reflect that at IFD?
- image-tiff does not support horizontal predictor for floating-point rasters, GDAL _does_. Since here the two layers are decoupled, I didn't put any checks, but just give the user bogus output if they give bogus input.
- output being Bytes, rather than having a `&mut [u8]` function input. The `&mut [u8]` would be preferred by me, because then it can be directly read into a user-provided buffer that has also ensured the alignment of the buffer (e.g. using initializing a `Vec<f32>` buffer and then bytemucking to `&mut [u8]`)

### DIFF
--- a/src/predictor.rs
+++ b/src/predictor.rs
@@ -1,0 +1,507 @@
+use std::collections::HashMap;
+use std::fmt::Debug;
+
+use byteorder::NativeEndian;
+use bytes::{Bytes, BytesMut};
+// use tiff::decoder::DecodingResult;
+
+use crate::{
+    error::AsyncTiffResult, reader::Endianness, tiff::tags::Predictor, tile::PredictorInfo,
+};
+
+/// A registry for reverse predictors
+///
+/// Reverse predictors, because they perform the inverse (decoding) operation of prediction
+///
+///
+///
+pub struct RevPredictorRegistry(HashMap<Predictor, Box<dyn RevPredictor>>);
+
+impl RevPredictorRegistry {
+    /// create a new predictor registry with no predictors registered
+    pub fn new() -> Self {
+        Self(HashMap::new())
+    }
+}
+
+impl AsRef<HashMap<Predictor, Box<dyn RevPredictor>>> for RevPredictorRegistry {
+    fn as_ref(&self) -> &HashMap<Predictor, Box<dyn RevPredictor>> {
+        &self.0
+    }
+}
+
+impl Default for RevPredictorRegistry {
+    fn default() -> Self {
+        let mut hmap = HashMap::new();
+        hmap.insert(Predictor::None, Box::new(NoPredictor) as _);
+        hmap.insert(Predictor::Horizontal, Box::new(RevHorizontalPredictor) as _);
+        hmap.insert(
+            Predictor::FloatingPoint,
+            Box::new(RevFloatingPointPredictor) as _,
+        );
+        Self(hmap)
+    }
+}
+
+/// Trait for reverse predictors to implement
+///
+///
+pub trait RevPredictor: Debug + Send + Sync {
+    /// reverse predict the decompressed bytes and fix endianness on the output
+    ///
+    ///
+    fn rev_predict_fix_endianness(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        tile_x: u32,
+        tile_y: u32,
+    ) -> AsyncTiffResult<Bytes>; // having this Bytes will give alignment issues later on
+}
+
+/// no predictor
+#[derive(Debug)]
+pub struct NoPredictor;
+
+impl RevPredictor for NoPredictor {
+    fn rev_predict_fix_endianness(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        _: u32,
+        _: u32,
+    ) -> AsyncTiffResult<Bytes> {
+        let mut res = BytesMut::from(buffer);
+        fix_endianness(
+            &mut res[..],
+            predictor_info.endianness,
+            predictor_info.bits_per_sample[0],
+        );
+        Ok(res.into())
+    }
+}
+
+/// reverse horizontal predictor
+#[derive(Debug)]
+pub struct RevHorizontalPredictor;
+
+impl RevPredictor for RevHorizontalPredictor {
+    fn rev_predict_fix_endianness(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        tile_x: u32,
+        _: u32,
+    ) -> AsyncTiffResult<Bytes> {
+        let output_row_stride = predictor_info.output_row_stride(tile_x)?;
+        let samples = predictor_info.samples_per_pixel as usize;
+        let bit_depth = predictor_info.bits_per_sample[0];
+
+        let mut res = BytesMut::from(buffer);
+        for buf in res.chunks_mut(output_row_stride) {
+            // this is rev_hpredict_nsamp from image-tiff
+            rev_hpredict_nsamp(buf, bit_depth, samples);
+            // end rev_hpredict_nsamp
+        }
+        fix_endianness(&mut res[..], predictor_info.endianness, bit_depth);
+        Ok(res.into())
+    }
+}
+
+/// Reverse predictor convenienve function for horizontal differencing
+///
+/// From image-tiff
+pub fn rev_hpredict_nsamp(buf: &mut [u8], bit_depth: u16, samples: usize) {
+    match bit_depth {
+        0..=8 => {
+            for i in samples..buf.len() {
+                buf[i] = buf[i].wrapping_add(buf[i - samples]);
+            }
+        }
+        9..=16 => {
+            for i in (samples * 2..buf.len()).step_by(2) {
+                let v = u16::from_ne_bytes(buf[i..][..2].try_into().unwrap());
+                let p = u16::from_ne_bytes(buf[i - 2 * samples..][..2].try_into().unwrap());
+                buf[i..][..2].copy_from_slice(&(v.wrapping_add(p)).to_ne_bytes());
+            }
+        }
+        17..=32 => {
+            for i in (samples * 4..buf.len()).step_by(4) {
+                let v = u32::from_ne_bytes(buf[i..][..4].try_into().unwrap());
+                let p = u32::from_ne_bytes(buf[i - 4 * samples..][..4].try_into().unwrap());
+                buf[i..][..4].copy_from_slice(&(v.wrapping_add(p)).to_ne_bytes());
+            }
+        }
+        33..=64 => {
+            for i in (samples * 8..buf.len()).step_by(8) {
+                let v = u64::from_ne_bytes(buf[i..][..8].try_into().unwrap());
+                let p = u64::from_ne_bytes(buf[i - 8 * samples..][..8].try_into().unwrap());
+                buf[i..][..8].copy_from_slice(&(v.wrapping_add(p)).to_ne_bytes());
+            }
+        }
+        _ => {
+            unreachable!("Caller should have validated arguments. Please file a bug.")
+        }
+    }
+}
+
+/// Fix endianness. If `byte_order` matches the host, then conversion is a no-op.
+///
+/// from image-tiff
+pub fn fix_endianness(buf: &mut [u8], byte_order: Endianness, bit_depth: u16) {
+    match byte_order {
+        Endianness::LittleEndian => match bit_depth {
+            0..=8 => {}
+            9..=16 => buf.chunks_exact_mut(2).for_each(|v| {
+                v.copy_from_slice(&u16::from_le_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+            17..=32 => buf.chunks_exact_mut(4).for_each(|v| {
+                v.copy_from_slice(&u32::from_le_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+            _ => buf.chunks_exact_mut(8).for_each(|v| {
+                v.copy_from_slice(&u64::from_le_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+        },
+        Endianness::BigEndian => match bit_depth {
+            0..=8 => {}
+            9..=16 => buf.chunks_exact_mut(2).for_each(|v| {
+                v.copy_from_slice(&u16::from_be_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+            17..=32 => buf.chunks_exact_mut(4).for_each(|v| {
+                v.copy_from_slice(&u32::from_be_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+            _ => buf.chunks_exact_mut(8).for_each(|v| {
+                v.copy_from_slice(&u64::from_be_bytes((*v).try_into().unwrap()).to_ne_bytes())
+            }),
+        },
+    };
+}
+
+/// Floating point predictor
+#[derive(Debug)]
+pub struct RevFloatingPointPredictor;
+
+impl RevPredictor for RevFloatingPointPredictor {
+    fn rev_predict_fix_endianness(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        tile_x: u32,
+        tile_y: u32,
+    ) -> AsyncTiffResult<Bytes> {
+        let output_row_stride = predictor_info.output_row_stride(tile_x)?;
+        let mut res: BytesMut = BytesMut::zeroed(
+            output_row_stride * predictor_info.chunk_height_pixels(tile_y)? as usize,
+        );
+        let bit_depth = predictor_info.bits_per_sample[0] as usize;
+        if predictor_info.chunk_width_pixels(tile_x)? == predictor_info.chunk_width {
+            // no special padding handling
+            let mut input = BytesMut::from(buffer);
+            for (in_buf, out_buf) in input
+                .chunks_mut(output_row_stride)
+                .zip(res.chunks_mut(output_row_stride))
+            {
+                match bit_depth {
+                    16 => rev_predict_f16(in_buf, out_buf, predictor_info.samples_per_pixel as _),
+                    32 => rev_predict_f32(in_buf, out_buf, predictor_info.samples_per_pixel as _),
+                    64 => rev_predict_f64(in_buf, out_buf, predictor_info.samples_per_pixel as _),
+                    _ => panic!("thou shalt not predict with float16"),
+                }
+            }
+        } else {
+            // specially handle padding bytes
+            // create a buffer for the full width
+            let mut input = BytesMut::from(buffer);
+
+            let input_row_stride =
+                predictor_info.chunk_width as usize * predictor_info.bits_per_pixel() / 8;
+            for (in_buf, out_buf) in input
+                .chunks_mut(input_row_stride)
+                .zip(res.chunks_mut(output_row_stride))
+            {
+                let mut out_row = BytesMut::zeroed(input_row_stride);
+                match bit_depth {
+                    16 => rev_predict_f16(in_buf, out_buf, predictor_info.samples_per_pixel as _),
+                    32 => {
+                        rev_predict_f32(in_buf, &mut out_row, predictor_info.samples_per_pixel as _)
+                    }
+                    64 => {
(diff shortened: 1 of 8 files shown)


src/tile.rs
```diff
@@ -2,26 +2,187 @@ use bytes::Bytes;
...
+
 /// A TIFF Tile response.
 ///
 /// This contains the required information to decode the tile. Decoding is separated from fetching
 /// so that sync and async operations can be separated and non-blocking.
 ///
 /// This is returned by `fetch_tile`.
 #[derive(Debug)]
-pub struct Tile {
+pub struct Tile<'a> {
```
kylebarron, line 174: Given that we have Python bindings, and we expose `Tile` to Python, we can't have a lifetime specifier here. It's not possible to have a `#[pyclass]` with internal lifetimes. 

Instead, we should remove any borrowed data from `Tile` and `PredictorInfo` and store `Arc<PredictorInfo>` here.

src/tile.rs
```diff
@@ -60,18 +221,69 @@ impl Tile {
     ///
     /// Decoding is separate from fetching so that sync and async operations do not block the same
     /// runtime.
-    pub fn decode(self, decoder_registry: &DecoderRegistry) -> AsyncTiffResult<Bytes> {
+    pub fn decode(
+        self,
+        decoder_registry: &DecoderRegistry,
+        predictor_registry: &RevPredictorRegistry,
```
kylebarron, line 227: For data decoding, we use a registry because:

- There are a large number of possible decompression algorithms, and I think some are still being added to the spec. It's not possible for us to ensure we _always_ cover all the decompressions that the end user might need.
- More decompressions could be added in the future
- Some decompressions may use bindings to native C code, and not be pure Rust. We wouldn't want to force non-Rust code onto any user.
- The user may want to opt-in to a different decompressor (e.g. a binding to C) for performance, so the registry allows them to override default behavior.

Do any of these criteria apply to predictors? The permissible values for the `Predictor` tag appear to be only:

- `None`
- `Horizontal`
- `FloatingPoint`

It seems like we could make this much simpler by not implementing it as a registry?

src/predictor.rs
```diff
@@ -0,0 +1,659 @@
...
+                buf[i..][..8].copy_from_slice(&(v.wrapping_add(p)).to_ne_bytes());
+            }
+        }
+        _ => {
+            unreachable!("Caller should have validated arguments. Please file a bug.")
+        }
+    }
+}
+
+/// Fix endianness. If `byte_order` matches the host, then conversion is a no-op.
```
kylebarron, line 148: It's unclear from reading this that does not copy data.

If it is truly intended to be a no-op, seems better to do 

```rs
    #[cfg(target_endian = "little")]
	Endianness::LittleEndian => {},
	#[cfg(target_endian = "big")]
	Endianness::LittleEndian => match bit_depth ...
```
Then it's more clear that nothing is happening when the endianness of the data matches the host system's endianness

src/predictor.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// reverse predict the decompressed bytes and fix endianness on the output
+    ///
+    ///
+    fn rev_predict_fix_endianness(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        tile_x: u32,
+        tile_y: u32,
+    ) -> AsyncTiffResult<Bytes>; // having this Bytes will give alignment issues later on
```
kylebarron, line 59: Why? `Bytes` is pretty much just `Arc<Vec<u8>>`

src/predictor.rs
```diff
@@ -0,0 +1,659 @@
...
+        let mut hmap = HashMap::new();
+        hmap.insert(Predictor::None, Box::new(NoPredictor) as _);
+        hmap.insert(Predictor::Horizontal, Box::new(RevHorizontalPredictor) as _);
+        hmap.insert(
+            Predictor::FloatingPoint,
+            Box::new(RevFloatingPointPredictor) as _,
+        );
+        Self(hmap)
+    }
+}
```
kylebarron, line 44: As mentioned previously, I think simplest to remove the registry concept from here. Then in `Tile::decode`, just match on the `Predictor` tag enum manually.

src/tile.rs
feefladder: I initially made this a registry because complexities wrt. `PlanarConfiguration` and/or multiple sample formats may be implemented by the user.
sidenode is that PlanarConfiguration `Planar` is rarely used, these [global CRS correction rasters](https://proj.org/en/stable/specifications/geodetictiffgrids.html) use `PlanarConfiguration::Planar` configuration?
I'm happy to ditch the registry and wait for someone to raise an issue on planarconfig
(note that image-tifff doesn't support `Planar` properly)

src/predictor.rs
feefladder: this was directly copied from `image-tiff`. will do!

src/tile.rs
```diff
@@ -2,26 +2,187 @@ use bytes::Bytes;
...
 
+/// All info that may be used by a predictor
+///
+/// Most of this is used by the floating point predictor
+/// since that intermixes padding into the decompressed output
+///
+/// Also provides convenience functions
+///
+#[derive(Debug, Clone, Copy)]
+pub struct PredictorInfo<'a> {
```
geospatial-jeff, line 20: I'm wondering if it makes sense to add this `PredictorInfo` struct because all of the information provided by it is already available on the IFD, and it seems like the main way of creating this struct is by calling `IFD.get_predictor_info`.

Any implementation that is unpredicting a tile will also have access to the IFD, so I think it could make sense to remove this intermediate struct entirely in favor of just interacting with the IFD directly.

src/tile.rs
```diff
@@ -2,26 +2,187 @@ use bytes::Bytes;
...
+                    assert_eq!(self.samples_per_pixel as usize, self.bits_per_sample.len());
+                    self.bits_per_sample.iter().map(|v| *v as usize).product()
+                }
+            }
+            PlanarConfiguration::Planar => self.bits_per_sample[0] as usize,
+        }
+    }
+
+    /// The number of chunks in the horizontal (x) direction
+    pub fn chunks_across(&self) -> u32 {
```
geospatial-jeff, line 157: To add to my previous comment on removing the `PredictorInfo` struct; both `chunks_across` and `chunks_down` are useful outside of handling predictors.  It would make sense to include these on the `IFD` struct (ex. `IFD.chunks_across`).

src/predictor.rs
feefladder: So if:
- the tiff is something with alignment>1, e.g. `f32`
- the global allocator gives out a misaligned Vec<u8> (which is doesn't often do)
then the user has two options:
1. copy over the data into a Vec<f32> using `f32::from_ne_bytes()`
2. use bytemuck and hope for the best
I looked into this quite deeply, and afaik, most "standard" allocators allocate with alignment 8, it would just save a copy in my mind?

kylebarron: I think instead we should discuss: at what points should be storing `Vec<u8>` and at what points be storing structured array types.

src/tile.rs
feefladder: `Tile` above says in its docstrings that it has all necessary information to decode a tile. I'm fine with removing the intermediate struct, but then I think it'd also make sense to not store `CompressionMethod` on the tile itself, since the decoding operation needs access to the IFD?

kylebarron: I actually disagree with @geospatial-jeff here. We created this `Tile` construct because we want to enable some use cases (including some of Jeff's) where the user doesn't actually want to decompress the data from the COG; they want to pass on the compressed data e.g. from a tile server. 

If we want to expose the `Tile` construct to bindings like Python, it can't have a lifetime. In that case, the `Tile` needs to maintain all data necessary to handle decoding, unprediction, and endianness checking. But the full IFD is quite large. So I think having a `PredictorInfo` struct is quite useful so that we can encapsulate only this data that we actually need to put in `Tile`. And it's useful that all of this `PredictorInfo` is also `Copy`.

src/predictor.rs
kylebarron: I think to add to my previous comment: the question is where we convert out of `Bytes`. I think the core networking trait (`AsyncFileReader`) should remain as it is, where `get_bytes` returns `bytes::Bytes`. Most networking code, at least [`reqwest`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes) and [`object_store`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get_range) return buffers as `Bytes`. 

There's no way to convert from a `Bytes` to a `Vec<u8>` zero-copy. (You can sometimes convert from a `Bytes` to `BytesMut` zero-copy). So that implies at _some point_ we make a data copy from a `Bytes` into a `Vec<T>`.

src/tile.rs
kylebarron: Even if we keep the `PredictorInfo`, as I think we probably should, we can put this functionality in a shared `TileMath` trait which both the IFD and `PredictorInfo` implement

src/tile.rs
```diff
@@ -2,26 +2,187 @@ use bytes::Bytes;
...
+    pub fn output_row_stride(&self, x: u32) -> AsyncTiffResult<usize> {
+        Ok((self.chunk_width_pixels(x)? as usize).saturating_mul(self.bits_per_pixel()) / 8)
+    }
+
+    /// The total number of bits per pixel, taking into account possible different sample sizes
+    ///
+    /// Technically bits_per_sample.len() should be *equal* to samples, but libtiff also allows
+    /// it to be a single value that applies to all samples.
+    ///
+    /// Libtiff and image-tiff do not support mixed bits per sample, but we give the possibility
```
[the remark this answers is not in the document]
kylebarron: Agreed we should only support consistent bits per sample, and then in the future we could relax that assumption if we need.

src/tile.rs
geospatial-jeff: I see, I didn't understand the limitation around "lifetime".  `PredictorInfo` makes sense to me if the goal is to have strong encapsulation around just the parts of the IFD that are needed for decoding.

kylebarron:
I made a PR with suggested changes into this branch. See https://github.com/feefladder/async-tiff/pull/1

feefladder:
Ah noo... I just finished incorporating feedback :'(

src/predictor.rs
feefladder: In contrast to `reqwest` and `object_store`, we actually know the underlying datatype

But maybe put this in a separate issue/PR? I didn't put it in here, because this PR was already medium-large and "my ideal situation^" would change some non-related things.

> I think to add to my previous comment: the question is where we convert out of `Bytes`. I think the core networking trait (`AsyncFileReader`) should remain as it is, where `get_bytes` returns `bytes::Bytes`. Most networking code, at least [`reqwest`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes) and [`object_store`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get_range) return buffers as `Bytes`.

I'm all up for not changing AsyncFileReader. I would say after the decompression step is where we can have typed arrays, since decompression itself is not zero-copy (except for no compression).

Then for the endianness fixing and horizontal predictor, it is also nice to have exclusive access to the underlying buffer (`&mut[u8]`), since that _is_ zero-copy (not the float predictor).

so something like:
```rust
// pseudocode
fn Tile.decode(&self, &decoder) -> AsyncTiffResult<DecodingResult> {
   let mut buffer: DecodingResult = todo!()
   let res: DecodingResult = decoder.decode(self.compressed_bytes, &mut buffer.cast());
   match self.predictor_info.predictor {
     // mutates in-place
     Predictor::Horizontal => unpredict_horizontal(res, &self.predictor_info, self.x)
     // also mutates in-place, but uses a copy
     Predictor::Float => unpredict_float(res, &self.predictor_info, self.x, self.y)
   }
}
```

> There's no way to convert from a `Bytes` to a `Vec<u8>` zero-copy. (You can sometimes convert from a `Bytes` to `BytesMut` zero-copy). So that implies at _some point_ we make a data copy from a `Bytes` into a `Vec<T>` in order to be safe. **Or** we could use similar code as from the Arrow project and build typed interfaces on top of an `Arc<Bytes>`, such as their [`Buffer`](https://github.com/apache/arrow-rs/blob/a35bdf0298acc5b4e06db166d44cbcdf7789edb0/arrow-buffer/src/buffer/immutable.rs#L68-L83) type.

In the current (this PR) implementation, I already did a `BytesMut::From()` quite some times, even though we were still in the same decoding step.

Sidenote: there is also some discussion going on over at image-tiff, where they want to just output a Bytes or &[u8], but then also the alignment issue was raised and somewhat ignored.

but... separate issue/PR?

src/predictor.rs
```diff
@@ -0,0 +1,784 @@
...
+    fn fix_endianness_and_unpredict(
+        &self,
+        buffer: Bytes,
+        predictor_info: &PredictorInfo,
+        tile_x: u32,
+        tile_y: u32,
+    ) -> AsyncTiffResult<Bytes> {
+        let output_row_stride = predictor_info.output_row_stride(tile_x)?;
+        let mut res: BytesMut =
+            BytesMut::zeroed(output_row_stride * predictor_info.output_rows(tile_y)?);
```
feefladder, line 294: Here now I did an output_rows, which is more if we have `Planar` planar config. I thought then at least we can give the user the data and they can split it themselves?

src/predictor.rs
```diff
@@ -0,0 +1,784 @@
...
+    }
+
+    fn bits_per_pixel(&self) -> usize {
+        match self.planar_configuration {
+            PlanarConfiguration::Chunky => {
+                self.bits_per_sample as usize * self.samples_per_pixel as usize
+            }
+            PlanarConfiguration::Planar => self.bits_per_sample as usize,
+        }
+    }
```
feefladder, line 125: these depend on planar_config

feefladder:
Thanks @kylebarron for the changes!

src/predictor.rs
```diff
@@ -0,0 +1,753 @@
...
+/// from image-tiff
+pub fn fix_endianness(buffer: Bytes, byte_order: Endianness, bit_depth: u16) -> Bytes {
+    #[cfg(target_endian = "little")]
+    if let Endianness::LittleEndian = byte_order {
+        return buffer;
+    }
+    #[cfg(target_endian = "big")]
+    if let Endianness::BigEndian = byte_order {
+        return buffer;
+    }
```
feefladder, line 216: I thought splitting the cfg->noop up here was a bit clearer

src/tile.rs
```diff
@@ -68,10 +71,24 @@ impl Tile {
...
+                decoded_tile,
+                self.predictor_info.endianness(),
+                self.predictor_info.bits_per_sample(),
+            )),
+            Predictor::Horizontal => {
+                unpredict_hdiff(decoded_tile, &self.predictor_info, self.x as _)
+            }
+            Predictor::FloatingPoint => {
+                unpredict_float(decoded_tile, &self.predictor_info, self.x as _, self.y as _)
+            }
```
feefladder, line 91: I removed the trait and only have crate-public functions now

src/tile.rs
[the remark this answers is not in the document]
feefladder: I think that point will never come and people that managed to get tiffs with mixed sample formats will be eternally sad that they can't use their files _anywhere_.

src/predictor.rs
feefladder: So now the `Planar` output will be:
```
[
Red...,
Green.,
Blue..,
]
```
where each band is `chunk_width_pixels()*chunk_height_pixels()`

src/tile.rs
feefladder: This should also help with less-copy as mocked up in #87

feefladder:
On another note: Since now there is this abstraction of "Chunks" for tiles and strips, the `Tile` struct doesn't make much sense anymore for stripped tiffs, but I think that's fine, and is merely a documentation issue

src/predictor.rs
feefladder: I thought this made sense, because the decoder (decompression) step decodes the entire data, which is like structured like this

src/predictor.rs
```diff
@@ -0,0 +1,753 @@
...
+/// floating point prediction first shuffles the bytes and then uses horizontal
+/// differencing
+/// also performs byte-order conversion if needed.
+pub fn rev_predict_f32(input: &mut [u8], output: &mut [u8], samples: usize) {
+    // reverse horizontal differencing
+    for i in samples..input.len() {
+        input[i] = input[i].wrapping_add(input[i - samples]);
+    }
+    // reverse byte shuffle and fix endianness
+    for (i, chunk) in output.chunks_mut(4).enumerate() {
```
weiji14, line 348: Would [`chunks_exact_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact_mut) be ok to use here? Is there a chance that chunks may not have exactly 4 bytes at the end?

src/predictor.rs
```diff
@@ -0,0 +1,753 @@
...
+        }
+        _ => {
+            unreachable!("Caller should have validated arguments. Please file a bug.")
+        }
+    }
+}
+
+/// Fix endianness. If `byte_order` matches the host, then conversion is a no-op.
+///
+/// from image-tiff
```
weiji14, line 207: Do you want this "from image-tiff" line to show up in the docs? Or only visible as a code comment?

```suggestion
// from image-tiff
```

src/predictor.rs
```diff
@@ -0,0 +1,753 @@
...
+    let mut res = BytesMut::from(buffer);
+    for buf in res.chunks_mut(output_row_stride) {
+        rev_hpredict_nsamp(buf, bit_depth, samples);
+    }
+    Ok(res.into())
+}
+
+/// Reverse predictor convenience function for horizontal differencing
+///
+/// From image-tiff
```
weiji14, line 168:
```suggestion
// From image-tiff
```

weiji14 (commented):
Code looks reasonable to me, only some minor suggestions around the docs, and one question I had around the reverse floating point prediction implementation.

src/predictor.rs
feefladder: should be ok, since the chunk-ness is based on `f32` being 4 bytes, also at the `f64` then

feefladder:
maybe ping @kylebarron who is/was main dev of this repo?

weiji14:
Oops, saw your message too late. Kyle's said he hasn't got much time for async-tiff lately, so he asked me to review. Feel free to open more patches as needed and I'll try my best to review them :smile:

feefladder:
ah ok great, I have one laying around for #87, lets see if that goes...

there was also a thingy around #82, where he put a lot of work in and I didn't appreciate (I was more like "why don't you merge my changes"). Turns out the direction he was going was actually better, (allowing for stateless reading and diversifying tag reading in downstream code) but #82 doesn't allow me to do, since some functions/structs I need are private. now since there were emotions involved and I think Github doesn't cater well to those types of conversations it all stalled a bit (I think also #83). I think all very solveable, but a bit sensitive I think? <- do you know a good way forward there?

but first lemme PR for #87

weiji14:
I'm not well versed enough in Rust to know whether these low-level implementation details matter, and we're all probably desiring slightly different things out of this library for very different use-cases (e.g. web-tiling, game-engines, gpu-shaders, etc). I say we can try and work on some low hanging fruit, and work on some core abstractions (e.g. I really want #100 to be solved), and anything super specific could live in our own personal repos like https://github.com/feefladder/tiff2, https://github.com/weiji14/cog3pio, etc. How does that sound?

feefladder:
Sounds good! tiff2 also implements all low-level stuff `async`ly, and I think that's better to do only in one place to prevent fragmentation of the ecosystem. Because I spent most time on tiff2, where the biggest chunk was low-level stuff, I don't really know that much about what would be needed for parsing geo tags or making a registry for that (what is needed for most tag parsing?). I could chip in on #100 a bit with some thoughts.

I already made a start at something called `async-tiff-wasm` (maybe async-cog-wasm is a better name) that would do the more specific stuff for reading cogs into wasm games etc. I think I'll open an issue about making #82 work for me when the need arises, for now I need to finish a report about these things.

PR #92, opened 2025-05-01
Enable pytest-asyncio tests in CI

weiji14:
## What I am changing

- Setup pytest-asyncio properly on GitHub Actions CI to run python unit tests.

## How I did it

- Enable `asyncio_mode="auto"` in the pytest config following https://pytest-asyncio.readthedocs.io/en/latest/reference/configuration.html#asyncio-mode
- Uncomment the `uv run pytest` line in `.github/workflows/test-python.yml`
- Refactor contents of `python/tests/test_cog.py` into a proper `async def test_cog_s3` function.

## How you can test it

- Run `cd python/ && uv run pytest --verbose` locally, or check the CI logs.

## Related Issues

-

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -60,6 +60,10 @@ jobs:
         with:
           version: "latest"
 
+      - name: uv sync
+        working-directory: python
+        run: uv sync --no-install-package geoindex-rs
+
       - name: maturin venv Build
         working-directory: python
         run: uv run maturin develop
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -60,10 +60,6 @@ jobs:
         with:
           version: "latest"
 
-      - name: uv sync
-        working-directory: python
-        run: uv sync --no-install-package geoindex-rs
-
       - name: maturin venv Build
         working-directory: python
         run: uv run maturin develop
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -56,10 +56,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install a specific version of uv
-        uses: astral-sh/setup-uv@v3
+        uses: astral-sh/setup-uv@v6
         with:
-          enable-cache: true
-          version: "0.5.x"
+          version: "latest"
 
       - name: uv sync
         working-directory: python
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: uv sync
         working-directory: python
-        run: uv sync --no-install-package geoindex-rs
+        run: uv sync --no-install-package async-tiff
 
       - name: maturin venv Build
         working-directory: python
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -67,8 +67,8 @@ jobs:
 
       - name: maturin venv Build
         working-directory: python
-        run: uv run --no-project maturin develop
+        run: uv run maturin develop
 
-      # - name: Run pytest
-      #   working-directory: python
-      #   run: uv run --no-project pytest
+      - name: Run pytest
+        working-directory: python
+        run: uv run pytest --verbose
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -66,8 +66,8 @@ jobs:
 
       - name: maturin venv Build
         working-directory: python
-        run: uv run maturin develop
+        run: uv run --no-project maturin develop
 
       - name: Run pytest
         working-directory: python
-        run: uv run pytest --verbose
+        run: uv run --no-project pytest --verbose
--- a/python/tests/test_cog.py
+++ b/python/tests/test_cog.py
@@ -1,5 +1,3 @@
-from time import time
-
 from async_tiff import TIFF, enums
 from async_tiff.store import S3Store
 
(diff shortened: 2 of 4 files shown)


.github/workflows/test-python.yml
```diff
@@ -67,8 +67,8 @@ jobs:
```
kylebarron, line 67: it looks like the line above this is incorrect

python/tests/test_cog.py
```diff
@@ -1,27 +1,31 @@
...
+    Ensure that TIFF.open can open a Sentinel-2 Cloud-Optimized GeoTIFF file from an
+    s3 bucket, read IFDs and GeoKeyDirectory metadata.
+    """
+    store = S3Store("sentinel-cogs", region="us-west-2", skip_signature=True)
+    path = "sentinel-s2-l2a-cogs/12/S/UF/2022/6/S2B_12SUF_20220609_0_L2A/B04.tif"
+
+    start = time()
+    tiff = await TIFF.open(path=path, store=store, prefetch=32768)
+    end = time()
+    assert end - start < 5  # take less than 5 seconds to open
```
kylebarron, line 18: I don't think we need to add this assert into the actual test. We can just take out the timing

.github/workflows/test-python.yml
```diff
@@ -67,8 +67,8 @@ jobs:
 
       - name: maturin venv Build
         working-directory: python
-        run: uv run --no-project maturin develop
```
kylebarron, line 70: The `--no-project` was here because I thought whenever we ran `uv sync` or `uv run`, uv would automatically build the current project, which would compile the rust code and take a while.

python/tests/test_cog.py
weiji14: Done in commit 28973b91a693c7b2fb6bfc9ba1d82020f5927e03

.github/workflows/test-python.yml
weiji14: You mean `uv sync --no-install-package geoindex-rs` should be just `uv sync`?

.github/workflows/test-python.yml
weiji14: Hmm, I though if the [working directory](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#working-directory) is set to `python/` already, then it should only compile what's inside the `python/` folder? Don't we need to compile the rust code anyway before compiling the pyo3 code? A little confused here.

.github/workflows/test-python.yml
```diff
@@ -56,19 +56,18 @@ jobs:
...
-        uses: astral-sh/setup-uv@v3
+        uses: astral-sh/setup-uv@v6
         with:
-          enable-cache: true
-          version: "0.5.x"
+          version: "latest"
 
       - name: uv sync
         working-directory: python
         run: uv sync --no-install-package geoindex-rs
```
weiji14, line 65: Actually, it should be possible to just remove the `uv sync ...` part, and just do `uv run maturin develop && uv run pytest` below?

.github/workflows/test-python.yml
kylebarron: no, it needs to be `uv sync --no-install-package async-tiff`. And then the lines after that still need the `--no-project`.

You see this line: https://github.com/developmentseed/async-tiff/actions/runs/14786441598/job/41515604193#step:7:19

When you run `uv sync` it _builds the rust code in release mode_ as part of the current project. For fastest CI we want to build and test in debug mode, not release mode. That means we want to tell `uv` **not** to automatically build `async-tiff`, because we'll manually build it after with `maturin develop`

.github/workflows/test-python.yml
kylebarron: It's not the cargo project that's the issue. It's the uv project. We don't want uv automatically building the rust parts of async-tiff in release mode.

.github/workflows/test-python.yml
weiji14: Ok, I think you're right. Checking the timings:

- Commit 12b230f57c67b7bb02105e3cc235d194c7c6d647, job https://github.com/developmentseed/async-tiff/actions/runs/14783290369/job/41506715402:
  - `uv sync` = 2m0s
  - `uv run maturin develop` = 46s
  - Total: 2m46s
- Commit 638d6029b03fe15262386ceb98015514ed8579d8, job https://github.com/developmentseed/async-tiff/actions/runs/14786316208/job/41515269407:
  - `uv run maturin develop` = 2m54s

So let's do `uv sync --no-install-package async-tiff` && `uv run --no-project maturin develop`.

weiji14: Final timings at commit aa213688c943406f8a2d9eaa6b4fcd7347857210, job https://github.com/developmentseed/async-tiff/actions/runs/14825301405/job/41617636468:
- `uv sync --no-install-package async-tiff` = 3s
- `uv run --no-project maturin develop` = 46s
- Total: 49s :rocket:

kylebarron (approved):
Thanks!

.github/workflows/test-python.yml
```diff
@@ -56,19 +56,18 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install a specific version of uv
-        uses: astral-sh/setup-uv@v3
+        uses: astral-sh/setup-uv@v6
         with:
-          enable-cache: true
```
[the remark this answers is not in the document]
weiji14: Yes, default was changed to `enable-cache: true` since setup-uv v5, see https://github.com/astral-sh/setup-uv/releases/tag/v5.0.0

PR #93, opened 2025-05-01
Raise FileNotFoundError instead of panic when opening missing files

### LINKED ISSUE #90: Raise NotFoundError rather than panic on non-existent file
A RustPanic occurs when passing a non-existent file to `TIFF.open()`. Would it be possible to instead raise a `FileNotFoundError` because this type of traceback would be unfamiliar to many Python users?

```Python
from async_tiff import TIFF
from async_tiff.store import LocalStore
import asyncio

file = "/Users/max/Documents/Code/maxrjones/virtual_tiff/tests/dvc/github/imaginary_file.tif" # Doesn't exist
store = LocalStore()
async def open_tiff(*, store, path):
    return await TIFF.open(path, store=store)

tiff = asyncio.run(open_tiff(store=store, path=file))
print(tiff)
```

```bash
(base) max@Maxs-DevSeed-MacBook python % uv run .vscode/no-tiff.py
thread 'tokio-runtime-worker' panicked at src/tiff.rs:35:86:
called `Result::unwrap()` on an `Err` value: ObjectStore(NotFound { path: "/Users/max/Documents/Code/maxrjones/virtual_tiff/tests/dvc/github/imaginary_file.tif", source: Os { code: 2, kind: NotFound, message: "No such file or directory" } })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/Users/max/Documents/Code/developmentseed/async-tiff/python/.vscode/no-tiff.py", line 11, in <module>
    tiff = asyncio.run(open_tiff(store=store, path=file))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/.local/share/uv/python/cpython-3.11.10-macos-aarch64-none/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/max/.local/share/uv/python/cpython-3.11.10-macos-aarch64-none/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/.local/share/uv/python/cpython-3.11.10-macos-aarch64-none/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/max/Documents/Code/developmentseed/async-tiff/python/.vscode/no-tiff.py", line 9, in open_tiff
    return await TIFF.open(path, store=store)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_async_runtimes.RustPanic: rust future panicked: unknown error
```

weiji14:
## What I am changing

- Avoid panic when non-existent file is passed into `TIFF.open`

## How I did it

- Replace `.unwrap()` with `.map_err(|err| PyFileNotFoundError::new_err(err.to_string()))?`
- Add unit test to ensure `FileNotFoundError` is raised for missing files

## How you can test it

- Run `cd python && uv run pytest --verbose` locally (preferrably on top of #92)

## Related Issues

Fixes #90

### DIFF
--- a/python/src/tiff.rs
+++ b/python/src/tiff.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use async_tiff::metadata::{PrefetchBuffer, TiffMetadataReader};
 use async_tiff::reader::AsyncFileReader;
 use async_tiff::TIFF;
-use pyo3::exceptions::PyIndexError;
+use pyo3::exceptions::{PyFileNotFoundError, PyIndexError};
 use pyo3::prelude::*;
 use pyo3::types::PyType;
 use pyo3_async_runtimes::tokio::future_into_py;
@@ -32,7 +32,9 @@ impl PyTIFF {
         let reader = store.into_async_file_reader(path);
 
         let cog_reader = future_into_py(py, async move {
-            let metadata_fetch = PrefetchBuffer::new(reader.clone(), prefetch).await.unwrap();
+            let metadata_fetch = PrefetchBuffer::new(reader.clone(), prefetch)
+                .await
+                .map_err(|err| PyFileNotFoundError::new_err(err.to_string()))?;
             let mut metadata_reader = TiffMetadataReader::try_open(&metadata_fetch).await.unwrap();
             let ifds = metadata_reader
                 .read_all_ifds(&metadata_fetch)
--- a/python/tests/test_cog.py
+++ b/python/tests/test_cog.py
@@ -1,7 +1,10 @@
 import async_tiff
 from time import time
+
+import pytest
+
 from async_tiff import TIFF
-from async_tiff.store import S3Store
+from async_tiff.store import LocalStore, S3Store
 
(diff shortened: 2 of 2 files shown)


python/src/tiff.rs
```diff
@@ -32,7 +32,9 @@ impl PyTIFF {
         let reader = store.into_async_file_reader(path);
 
         let cog_reader = future_into_py(py, async move {
-            let metadata_fetch = PrefetchBuffer::new(reader.clone(), prefetch).await.unwrap();
```
kylebarron, line 35: Ideally we should remove all of these `unwrap`s, but we can leave that for the future.

kylebarron (approved):

PR #99, opened 2025-05-13
Raise TypeError instead of panic on doing fetch_tile from striped TIFFs

### LINKED ISSUE #98: `rust future panicked: unknown error` when TIFF contains strips rather than tiles
`TIFF.fetch_tile()` panics if the TIFF contains strips defined by `IFD.rows_per_strip` rather than tiles defined by `IFD.tile_width`, `IFD.tile_height`, etc.

MVCE:
```Python
import obstore as obs
from async_tiff import TIFF
from async_tiff.store import LocalStore
import os
import asyncio


# Set which file to run the test on
file = "int8.tif"
# Download file
store = obs.store.HTTPStore.from_url("https://dagshub.com")

path = f"maxrjones/virtual-tiff/raw/73934fbfb31cd5bb099f225a47ac8ea34298f3ea/tests/dvc/gdal_autotest/{file}"

resp = obs.get(store, path)
with open(file, "wb") as f:
    for chunk in resp:
        f.write(chunk)

async def load_tile(*, store, path):
    tiff = await TIFF.open(path, store=store)
    return await tiff.fetch_tile(0, 0, 0)

store = LocalStore()
filepath = f"{os.getcwd()}/{file}"

tile = asyncio.run(load_tile(store=store, path=filepath))
print(tile)
```

weiji14:
## What I am changing

- Avoid panic when running `.fetch_tile()` on a striped TIFF (that does not have tiles).

## How I did it

- Replace `.unwrap()` with `.map_err(|err| PyTypeError::new_err(err.to_string()))`
- Add unit test to ensure `TypeError` is raised when calling `.fetch_tile()` on a striped TIFF

## How you can test it

- Run `cd python && uv run pytest --verbose` or look at CI

## Related Issues

Closes #98

### DIFF
--- a/python/src/tiff.rs
+++ b/python/src/tiff.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use async_tiff::metadata::{PrefetchBuffer, TiffMetadataReader};
 use async_tiff::reader::AsyncFileReader;
 use async_tiff::TIFF;
-use pyo3::exceptions::{PyFileNotFoundError, PyIndexError};
+use pyo3::exceptions::{PyFileNotFoundError, PyIndexError, PyTypeError};
 use pyo3::prelude::*;
 use pyo3::types::PyType;
 use pyo3_async_runtimes::tokio::future_into_py;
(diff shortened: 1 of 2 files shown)


python/src/tiff.rs
```diff
@@ -69,7 +69,11 @@ impl PyTIFF {
             // TODO: avoid this clone; add Arc to underlying rust code?
             .clone();
         future_into_py(py, async move {
-            let tile = ifd.fetch_tile(x, y, reader.as_ref()).await.unwrap();
+            let tile = ifd
+                .fetch_tile(x, y, reader.as_ref())
+                .await
+                .map_err(|err| PyTypeError::new_err(err.to_string()))?;
+
```
kylebarron, line 76: Ideally we'd have an error enum with known variants so that we can check specifically for "trying to load a strip tiff as a tiled tiff" and return a Python TypeError for that specific case.

But this is fine for now.

kylebarron (approved):

Issue #101, opened 2025-05-19
OME-TIFF: `Result::unwrap()` on an `Err` value: General("invalid utf-8 sequence of 1 bytes from index 2144")


weiji14:
> FYI I decided to try the OME-TIFF file from https://www.cellimagelibrary.org/images/40613 and encountered another panic on opening (xref https://github.com/virtual-zarr/virtual-tiff/issues/32). 

 _Originally posted by @maxrjones in [#97](https://github.com/developmentseed/async-tiff/issues/97#issuecomment-2887813181)_

Code to reproduce panic:

```python
import asyncio
from async_tiff.store import HTTPStore

from async_tiff import TIFF

async def test_ome_tiff():
    store = HTTPStore(url="https://cildata.crbs.ucsd.edu/")
    path = "media/images/40613/40613.tif"

    tiff = await TIFF.open(path=path, store=store)
    return tiff

if __name__ == "__main__":
    asyncio.run(test_ome_tiff())
```

Full traceback:

```python-traceback
thread 'tokio-runtime-worker' panicked at python/src/tiff.rs:44:18:
called `Result::unwrap()` on an `Err` value: General("invalid utf-8 sequence of 1 bytes from index 2144")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/home/user/projects/async-tiff/python/tests/test_ome_tiff.py", line 58, in <module>
    asyncio.run(test_ome_tiff())
  File "/home/user/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/user/Documents/github/async-tiff/python/tests/test_ome_tiff.py", line 49, in test_ome_tiff
    tiff = await TIFF.open(path=path, store=store)
[...]

PR #102, opened 2025-05-21
Test opening single-channel OME-TIFF file

weiji14:
## What I am changing

- Ensure that a single-channel OME-TIFF files can be opened by `async-tiff`

## How I did it

- Created a helper `open_remote_tiff` utility function to read TIFFs from object stores
- Added a test to open an OME-TIFF file from https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/single-channel.ome.tif, ensuring that the ImageDescription [OME XML metadata](https://ome-model.readthedocs.io/en/stable/ome-tiff/specification.html#ome-xml-metadata) is parsed out correctly.

## How you can test it

- Run `cargo test` locally or look at CI logs

## Related Issues

- #101

### DIFF
--- a/tests/ome_tiff.rs
+++ b/tests/ome_tiff.rs
@@ -1,0 +1,21 @@
+/// Integration tests on OME-TIFF files.
+use async_tiff::tiff::tags::PhotometricInterpretation;
+
+mod util;
+
+#[tokio::test]
+async fn test_ome_tiff_single_channel() {
+    let tiff =
+        util::open_remote_tiff("https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/single-channel.ome.tif").await;
+
+    assert_eq!(tiff.ifds().len(), 1);
+    let ifd = &tiff.ifds()[0];
+
+    assert_eq!(
+        ifd.photometric_interpretation(),
+        PhotometricInterpretation::BlackIsZero
+    );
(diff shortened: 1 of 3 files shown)


kylebarron:
Does this close #101?

tests/ome_tiff.rs
```diff
@@ -0,0 +1,21 @@
+/// Integration tests on OME-TIFF files.
+use async_tiff::tiff::tags::PhotometricInterpretation;
+
+mod util;
+
+#[tokio::test]
+async fn test_ome_tiff_single_channel() {
+    let tiff =
+        util::open_remote_tiff("https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/single-channel.ome.tif").await;
```
kylebarron, line 9: yeah I think it would be better to either check in this file into this repo, or perhaps create an `async-tiff-fixtures` separate repo that we can check data into.

Assuming this file is large, it would be great if we could extract like a one-pixel or very small subset of this file, while maintaining all header metadata the same.

_Or_, we could find the size of the last IFD and just clip the existing buffer to that region. So all IFD parsing would succeed but any image loading would fail as the image data is past the byte region checked in to the test fixture.

weiji14:
> Does this close #101?

No it doesn't. I haven't applied the `String::from_utf8_lossy` patch mentioned at https://github.com/developmentseed/async-tiff/issues/101#issuecomment-2896574826, nor handled the line causing the panic on the pyo3 side yet. Will do that in another PR.

kylebarron:
Thanks

PR #103, opened 2025-06-12
added `DecodingResult`  struct, changed decoding and predictor API to allow in-place 

### LINKED ISSUE #87: Where to go from `Bytes` to `DecodingResult`?
This emerged in #86 and is important, but may need some more discussion/other changes. See [there](https://github.com/developmentseed/async-tiff/pull/86#discussion_r2025453818) for more info.

feefladder:
closes #87 also #86 is somewhat relevant.

makes the output of a decoding operation a typed array.

The goals:
- output data is properly aligned and typed
- less-copy: try to copy the resulting data as little as possible
the problem:
1. data comes in as Bytes from the underlying network layer -> 1 copy is needed for typing+alignment (without extensive spec-reading and changing the api even more if we ever wanted to directly cast the Bytes to aligned+typed arrays)
2. some decompressions allow for reading into a buffer (deflate, lzw), others (~lzw~, jpeg) not. uncompressed copies because of 1
3. some predictors allow in-place (none: endianness-fixing and horizontal), others (float) not

The typed array can be filled in different places:
1. during decoding (breaking change, this PR)
2. during unpredicting (less breaking, leaves DecoderRegistry API untouched)
3. after unpredicting (current, the user does that here)

this PR does it during the decoding step, which:
- 6allows decompression algorithms to directly decompress into the buffer, saving a copy
- allows users to create their own typed buffers and decode into those using `decode_into`.

It changes quite some public api, so open to discussion. more comments coming on the changes

sidenote: there is a redundant copy with no compression and floating point prediction, but which sensible person would ever have that configuration?

### DIFF
--- a/src/bytecast.rs
+++ b/src/bytecast.rs
@@ -1,0 +1,36 @@
+//! Trivial, internal byte transmutation.
+//!
+//! A dependency like bytemuck would give us extra assurance of the safety but overall would not
+//! reduce the amount of total unsafety. We don't use it in the interface where the traits would
+//! really become useful.
+//!
+//! SAFETY: These are benign casts as we apply them to fixed size integer types only. All of them
+//! are naturally aligned, valid for all bit patterns and their alignment is surely at most their
+//! size (we assert the latter fact since it is 'implementation defined' if following the letter of
+//! the unsafe code guidelines).
+//!
+//! TODO: Would like to use std-lib here.
+// from image-tiff src/bytecast.rs
+// see https://stackoverflow.com/a/29042896/14681457
+use std::{mem, slice};
+
+macro_rules! integral_slice_as_bytes{($int:ty, $const:ident $(,$mut:ident)*) => {
+    pub(crate) fn $const(slice: &[$int]) -> &[u8] {
+        assert!(mem::align_of::<$int>() <= mem::size_of::<$int>());
+        unsafe { slice::from_raw_parts(slice.as_ptr() as *const u8, mem::size_of_val(slice)) }
+    }
+    $(pub(crate) fn $mut(slice: &mut [$int]) -> &mut [u8] {
+        assert!(mem::align_of::<$int>() <= mem::size_of::<$int>());
+        unsafe { slice::from_raw_parts_mut(slice.as_mut_ptr() as *mut u8, mem::size_of_val(slice)) }
+    })*
+}}
+
+integral_slice_as_bytes!(i8, i8_as_ne_bytes, i8_as_ne_mut_bytes);
+integral_slice_as_bytes!(u16, u16_as_ne_bytes, u16_as_ne_mut_bytes);
+integral_slice_as_bytes!(i16, i16_as_ne_bytes, i16_as_ne_mut_bytes);
+integral_slice_as_bytes!(u32, u32_as_ne_bytes, u32_as_ne_mut_bytes);
+integral_slice_as_bytes!(i32, i32_as_ne_bytes, i32_as_ne_mut_bytes);
+integral_slice_as_bytes!(u64, u64_as_ne_bytes, u64_as_ne_mut_bytes);
+integral_slice_as_bytes!(i64, i64_as_ne_bytes, i64_as_ne_mut_bytes);
+integral_slice_as_bytes!(f32, f32_as_ne_bytes, f32_as_ne_mut_bytes);
+integral_slice_as_bytes!(f64, f64_as_ne_bytes, f64_as_ne_mut_bytes);
--- a/src/cog.rs
+++ b/src/cog.rs
@@ -52,7 +52,7 @@ mod test {
         let ifd = &tiff.ifds[1];
         let tile = ifd.fetch_tile(0, 0, reader.as_ref()).await.unwrap();
         let tile = tile.decode(&Default::default()).unwrap();
-        std::fs::write("img.buf", tile).unwrap();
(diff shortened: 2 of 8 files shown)


feefladder:
oof it seems putting the change at the decoder step introduces some Python headaches. [this pyo3 issue](https://github.com/PyO3/pyo3/issues/1385) seems relevant. Mainly that `result_buffer: &mut[u8]` which is now passed to the decoder(s) is not python-friendly.... have to go now

feefladder:
Made it a draft, because currently the Python side would need changes as well and I'm as of yet inexperienced with Rust-Python things. Can look into it further after wednesday. For the moment, any thoughts @weiji14? especially if changing to typed array at the unpredicting step is the way to go? (would be slightly sad because of the extra copy, but that's ok and a good compromise I think - if the python stuff turns out to be a lot/difficult or too breaking)

alternatively, this PR could be split into:
- one adding the DecodingResult enum, creation logic and its Pythonic equivalents
- one that changes the actual decoding steps

weiji14:
> alternatively, this PR could be split into:
> 
>     * one adding the DecodingResult enum, creation logic and its Pythonic equivalents
>     * one that changes the actual decoding steps

Yes please, might be good to split this into two (I'm generally in favour of reviewing smaller PRs). I'm wondering if there's an opportunity ot use `num-trait`'s [`ToBytes`](https://docs.rs/num-traits/0.2.19/num_traits/ops/bytes/trait.ToBytes.html) trait here somewhere, but haven't thought about it too much yet.

feefladder:
> > alternatively, this PR could be split into:
> > ```
> > * one adding the DecodingResult enum, creation logic and its Pythonic equivalents
> > * one that changes the actual decoding steps
> > ```
> 
> Yes please, might be good to split this into two (I'm generally in favour of reviewing smaller PRs).

will do! I'll keep it in one step for now, because most changes are coherent and basically I won't know what is needed before I see things working together. So the plan:

1. make everything work
2. split it in two
3. add nice unit tests
4. make non-draft PRs

> I'm wondering if there's an opportunity ot use `num-trait`'s [`ToBytes`](https://docs.rs/num-traits/0.2.19/num_traits/ops/bytes/trait.ToBytes.html) trait here somewhere, but haven't thought about it too much yet.

Is `num-trait`'s `ToBytes` `Bytes` the same as `bytes::Bytes`? If so that would be good

I've given the whole Python part a bit more thought. passing a `&mut[u8]` across the boundary adds a lot of caveats on the python side or requires implementing the buffer protocol for ~`DecodingResult`~ `RasterData` (I think that's a better name), which is IMO a no-go (the added complexity is absolutely not worth it). Should be as simple as:
```rust
impl Decoder for PyDecoder {
    fn decode_tile(
        &self,
        compressed_buffer: Bytes,
        result_buffer: &mut[u8],
        _photometric_interpretation: PhotometricInterpretation,
        _jpeg_tables: Option<&[u8]>,
    ) -> AsyncTiffResult<()> {
        let decoded_buffer = Python::with_gil(|py| self.call(py, buffer))
            .map_err(|err| AsyncTiffError::General(err.to_string()))?;
        result_buffer.copy_from_slice(&decoded_buffer.into_inner()); //only add this extra copy
        Ok(())
    }
}
```
then there is a bit of sadness (needless copy) in the case of a user-implemented decoder and floating point predictor, but solving that would also be unnecessarily ugly for saving only a copy.

kylebarron:
> I'm wondering if there's an opportunity ot use `num-trait`'s [`ToBytes`](https://docs.rs/num-traits/0.2.19/num_traits/ops/bytes/trait.ToBytes.html) trait here somewhere

No, that `ToBytes` is for converting a single number to a byte array. Like converting a single u32 to the four bytes that make it up. So it's irrelevant for our needs I think.

kylebarron:
> oof it seems putting the change at the decoder step introduces some Python headaches. [this pyo3 issue](https://github.com/PyO3/pyo3/issues/1385) seems relevant. Mainly that `result_buffer: &mut[u8]` which is now passed to the decoder(s) is not python-friendly.... have to go now

This is another reason why we should use `bytes::Bytes` everywhere, so that we can easily reuse free buffer protocol interop with https://docs.rs/pyo3-bytes/0.3.0/pyo3_bytes/

feefladder:
> > oof it seems putting the change at the decoder step introduces some Python headaches. [this pyo3 issue](https://github.com/PyO3/pyo3/issues/1385) seems relevant. Mainly that `result_buffer: &mut[u8]` which is now passed to the decoder(s) is not python-friendly.... have to go now
> 
> This is another reason why we should use `bytes::Bytes` everywhere, so that we can easily reuse free buffer protocol interop with https://docs.rs/pyo3-bytes/0.3.0/pyo3_bytes/

So my final snippet did not change anything Python-API side and solved the issue I raised here. To keep everything Bytes, one would:
```rust
impl Decoder for PyDecoder {
    fn decode_tile(
        &self,
        compressed_buffer: Bytes,
        result_buffer: &mut[u8],
        _photometric_interpretation: PhotometricInterpretation,
        _jpeg_tables: Option<&[u8]>,
    ) -> AsyncTiffResult<()> {
        let decoded_buffer = Python::with_gil(|py| self.call(py, buffer))
            .map_err(|err| AsyncTiffError::General(err.to_string()))?;
        result_buffer.copy_from_slice(&decoded_buffer.into_inner()); //only add this extra copy
        Ok(())
    }
}
```
...I guess it could indeed have some more thought

Issue #69, opened 2025-03-18
Reorder repo layout; remove rust project from top level?


kylebarron:
We should be able to follow the obstore repo layout more closely, which may be less confusing to jump between.

kylebarron:
I think this depends on whether you consider the library to be more "rust-first" or "python-first". In the case of [obstore](https://github.com/developmentseed/obstore) the repo is more "python-first", while in the case of [geo-index](https://github.com/georust/geo-index) the repo is more "rust-first" with the addition of Python bindings.

I think for now, it's ok to leave the repo as it is, to more or less leave it as a "rust-first" library with Python bindings.

PR #106, opened 2025-06-14
added Endianness access from IFD to python

### LINKED ISSUE #91: Add endianness property to the TIFF class
It would be helpful to know the endianness of the file for building https://github.com/maxrjones/virtual-tiff. While I could get the first 2 bytes of the file using a separate file reader, it'd be convenient to get this from a property on the TIFF class. Would this be in scope for `async_tiff`?

feefladder:
closes #91 No unit tests yet. There were none before me in the Python part. Would be happy to add some with some pointers.

I chose to duplicate the enum in Python, so users can access it as `Endianness.LittleEndian` in Python as well, using pyo3 macro derivation.

I added the repr(u16) and custom discriminant on `PyEndianness` for matching tiff structure. I could reflect that in src/reader.rs::`Endianness` or drop completely.

It seems that `FromPyObject` trait impl is needed if we want to pass it into a rust function from the python side. Since we don't currently do that, I didn't implement it. Could be nicely done to accept all of `Endianness`, `u16` or `String` I think? I think in that case we need to also register the class in `python/src/lib.rs::_async_tiff` so they can actually create it in python. - I think that's needed anyways... will commit

I'm a bit unsure of the location of `PyEndianness` Since its an enum I put it in python/src/`enums.rs`, even though all enums there are tags. On the rust side `Endianness` is in `reader.rs`, which makes sense implementation-wise I think? What would be a good place? Or should we wait for repo layout (#69).

### DIFF
--- a/python/src/enums.rs
+++ b/python/src/enums.rs
@@ -1,10 +1,39 @@
-use async_tiff::tiff::tags::{
-    CompressionMethod, PhotometricInterpretation, PlanarConfiguration, Predictor, ResolutionUnit,
-    SampleFormat,
+use async_tiff::{
+    reader::Endianness,
+    tiff::tags::{
+        CompressionMethod, PhotometricInterpretation, PlanarConfiguration, Predictor,
+        ResolutionUnit, SampleFormat,
+    },
 };
 use pyo3::prelude::*;
 use pyo3::types::{PyString, PyTuple};
 use pyo3::{intern, IntoPyObjectExt};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[pyclass(eq, eq_int, name = "Endianness")]
+#[repr(u16)]
+pub(crate) enum PyEndianness {
+    LittleEndian = 0x4949, // b"II"
+    BigEndian = 0x4D4D,    // b"MM"
+}
+
+impl From<Endianness> for PyEndianness {
+    fn from(value: Endianness) -> Self {
+        match value {
+            Endianness::LittleEndian => Self::LittleEndian,
+            Endianness::BigEndian => Self::BigEndian,
+        }
+    }
+}
+
+impl From<PyEndianness> for Endianness {
+    fn from(value: PyEndianness) -> Self {
+        match value {
+            PyEndianness::LittleEndian => Self::LittleEndian,
+            PyEndianness::BigEndian => Self::BigEndian,
+        }
+    }
+}
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub(crate) struct PyCompressionMethod(CompressionMethod);
--- a/python/src/ifd.rs
+++ b/python/src/ifd.rs
@@ -4,8 +4,8 @@ use async_tiff::ImageFileDirectory;
 use pyo3::prelude::*;
 
 use crate::enums::{
-    PyCompressionMethod, PyPhotometricInterpretation, PyPlanarConfiguration, PyPredictor,
-    PyResolutionUnit, PySampleFormat,
+    PyCompressionMethod, PyEndianness, PyPhotometricInterpretation, PyPlanarConfiguration,
+    PyPredictor, PyResolutionUnit, PySampleFormat,
 };
 use crate::geo::PyGeoKeyDirectory;
 use crate::value::PyValue;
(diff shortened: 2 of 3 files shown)


kylebarron:
You can see my preferred way of implementing pyo3 enums here:

- pure-python enum: https://github.com/geoarrow/geoarrow-rs/blob/c8561fb61bd21a4455a7b7741bf80a6935ae6060/python/geoarrow-core/python/geoarrow/rust/core/enums.py#L9-L22
- `FromPyObject` that checks string values https://github.com/geoarrow/geoarrow-rs/blob/c8561fb61bd21a4455a7b7741bf80a6935ae6060/rust/pyo3-geoarrow/src/coord_type.rs#L13-L22 (although here you shouldn't need to implement `FromPyObject` at all because the user never passes in endianness)
- `IntoPyObject` that returns the Python enum instance: https://github.com/geoarrow/geoarrow-rs/blob/c8561fb61bd21a4455a7b7741bf80a6935ae6060/rust/pyo3-geoarrow/src/coord_type.rs#L42-L55

python/src/enums.rs
```diff
@@ -1,11 +1,46 @@
-use async_tiff::tiff::tags::{
-    CompressionMethod, PhotometricInterpretation, PlanarConfiguration, Predictor, ResolutionUnit,
-    SampleFormat,
+use async_tiff::{
+    reader::Endianness,
+    tiff::tags::{
+        CompressionMethod, PhotometricInterpretation, PlanarConfiguration, Predictor,
+        ResolutionUnit, SampleFormat,
+    },
```
kylebarron, line 6: Can we keep this as module imports instead of crate imports.

python/python/async_tiff/enums.py
```diff
@@ -1,5 +1,22 @@
...
 
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from enum import Enum
+    class StrEnum(str, Enum):
+        def __str__(self):
+            return str(self.value)
+
+class Endianness(StrEnum):
```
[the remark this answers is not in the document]
feefladder: I see you opened an issue for that. I could pull it into this PR though?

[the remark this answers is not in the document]
kylebarron: I'd prefer a separate PR anyways. You can look at https://github.com/developmentseed/obstore/blob/ec7dd298f62bf48a23dd677d3780540ece883bfb/.pre-commit-config.yaml#L20-L25 and https://github.com/developmentseed/obstore/blob/ec7dd298f62bf48a23dd677d3780540ece883bfb/.github/workflows/test-python.yml#L17-L39

python/python/async_tiff/_ifd.pyi
```diff
@@ -11,6 +12,8 @@ from ._geo import GeoKeyDirectory
 Value = int | float | str | tuple[int, int] | list[Value]
 
 class ImageFileDirectory:
+    @property
+    def endianness(self) -> Endianness: ...
```
feefladder, line 16: I did that

python/python/async_tiff/_tile.pyi
```diff
@@ -6,18 +6,23 @@ from ._thread_pool import ThreadPool
...
         """The row index this tile represents."""
+
     @property
     def compressed_bytes(self) -> Buffer:
         """The compressed bytes underlying this tile."""
+
     @property
     def compression_method(self) -> CompressionMethod | int:
         """The compression method used by this tile."""
+
```
feefladder, line 25: black added these

python/src/ifd.rs
```diff
@@ -15,6 +15,11 @@ pub(crate) struct PyImageFileDirectory(ImageFileDirectory);
 
 #[pymethods]
 impl PyImageFileDirectory {
+    #[getter]
+    pub fn endianness(&self) -> PyEndianness {
+        self.0.endianness().into()
+    }
```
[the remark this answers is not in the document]
feefladder: I did that (lines 15-16, also comment). What I meant is a `-> Endianness`

python/python/async_tiff/_tile.pyi
kylebarron: In any case, I'd prefer for ruff format to be added as a separate PR with CI validation first, just like the cargo import ordering

PR #112, opened 2025-07-29
feat:pre-commit Added pre-commit file and added to Python CI

### LINKED ISSUE #108: We need to add Python formatting to CI; as this would fail `black`/`ruff fmt`.
We need to add Python formatting to CI; as this would fail `black`/`ruff fmt`.

_Originally posted by @kylebarron in https://github.com/developmentseed/async-tiff/pull/106#discussion_r2152452136_

feefladder:
closes #108 

I also added rust pre-commit hooks that run on pre-push, since that prevents CI run failures I often encounter, while they should still not run in python CI (let's see when this PR does its CI)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,50 @@
...
+        stages: [pre-push]
+        pass_filenames: false
+
+      - id: rust-test
+        name: cargo test
+        entry: cargo test --all --all-features
+        language: system
+        types: [rust]
+        stages: [pre-push]
+        pass_filenames: false
```
feefladder, line 50: Since these stages are pre-push, they don't run in CI [as can be seen here](https://github.com/developmentseed/async-tiff/actions/runs/16605464880/job/46976011281#step:6:46)

python/python/async_tiff/__init__.py
```diff
@@ -1,9 +1,9 @@
 from typing import TYPE_CHECKING
 
-from ._async_tiff import *
+from ._async_tiff import *  # noqa: F403
 from ._async_tiff import ___version
 
 if TYPE_CHECKING:
-    from . import store
+    from . import store  # noqa: F401
```
feefladder, line 7: This gave ruff errors, not sure if we want to fix those or ignore like now

tests/image_tiff/images/COPYRIGHT
```diff
@@ -1,21 +1,21 @@
...
+EXPRESS, IMPLIED OR OTHERWISE, INCLUDING WITHOUT LIMITATION, ANY
+WARRANTY OF MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.
 
 IN NO EVENT SHALL SAM LEFFLER OR SILICON GRAPHICS BE LIABLE FOR
 ANY SPECIAL, INCIDENTAL, INDIRECT OR CONSEQUENTIAL DAMAGES OF ANY KIND,
 OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
-WHETHER OR NOT ADVISED OF THE POSSIBILITY OF DAMAGE, AND ON ANY THEORY OF 
-LIABILITY, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE 
+WHETHER OR NOT ADVISED OF THE POSSIBILITY OF DAMAGE, AND ON ANY THEORY OF
+LIABILITY, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
```
feefladder, line 20: trim-whitespace

.pre-commit-config.yaml
kylebarron: I'm lukewarm about using pre-commit to manage Rust linting. pre-commit is mostly just used in the Python community IIRC

feefladder: hmm... there was [a repo](https://github.com/doublify/pre-commit-rust/issues) for rust-like linting, but it was rather stale, so I didn't use that.

I searched a bit and found this [commitizen blog](https://rodneylab.com/rust-ci-tooling/#pre-commit-config) where they use it also for rust stuff. Whether this commitizen thing is wanted here I don't know. also [this rather large project using pre-commit](https://github.com/bpfman/bpfman/blob/2ce46c8d8fa32f842259badd10c86f2250859a64/.pre-commit-config.yaml#L13) almost the same as here

Then some rust-specific pre-commit info directly [uses](https://deaddabe.fr/blog/2021/09/29/git-pre-commit-hook-for-rust-projects/) [custom scripts](https://stackoverflow.com/a/78397712/14681457) (which I think is suboptimal)

I understand your lukewarm support though, it's a bit double to have it both in pre-commit and CI (and those going out of sync). However, I always spend wayy too much time having failing fmt/clippy tests in CI, so added it here, since I was working on it anyways.

feefladder: I now added a comment to clarify that they do nothing unless explicitly installed through `pre-commit install -t pre-push`
I thought that way it's a clear opt-in for people like me?

PR #113, opened 2025-07-29
chore:clippy

feefladder:
#112 had unrelated clippy errors due to clippy having been updated. Those are fixed in this PR.


kylebarron (approved):
Yes, 1.88 added new clippy lints; thanks!

PR #114, opened 2025-07-30
feat:extra_tags_registry

### LINKED ISSUE #100: Extension tag registry
> Just one quick note: I wonder if we could have a registry of extension tags. I'd love to remove geo specific tags from this repo, or at least remove them from their special status. Perhaps we could let users register custom tag codes plus how to parse them? 

 _Originally posted by @kylebarron in [#97](https://github.com/developmentseed/async-tiff/issues/97#issuecomment-2874995986)_

feefladder:
closes #100

This PR only does the registry and the needed API. There should be a separate PR coming that moves all geo-tag parsing code to use this system. I thought there'd be enough changes in this PR already

notes:
- deep clone functionality ?needed? (since different IFDs can have different tag values)
  - this uses some pointer comparison magic where I cast to "thin" pointers, which technically shouldn't matter, since we're only expecting to have duplicates from `Arc::clone`s, but it's a bit of a large possible footgun that's easily avoided.
  - there is a bit of a dance to make `Clone` object-safe, so there's the blanket `ExtraTagsCloneArc` trait.
- made the API match that of DecoderRegistry: pass in the registry as an argument to the last possible function call (`IFDReader::read`;`IFDReader::read_all_ifds()`)
- The trait `fn tags(&self) -> &'static [Tag]` is a bit ugly, also in the tests
  - `&self` is needed because of object safety
  - should return a reference to a `static` or `const` variable 
- more tests still desireable for `ifd::from_tags`
  - this is now in exif example doctests
  - deep cloning in `read_all_tags` is not currently tested
- keep the same functionality in the later PR by adding `GeoTags` to `ExtraTagsRegistry::default()`
- could have a non-mutable access function to the underlying hashmap

@weiji14 you gave 👀, feel like reviewing? :)


weiji14:
Hi @feefladder, I'm super busy right now (upcoming conference + travel), so won't have the headspace to review until late August. Maybe @kylebarron can have a look if he has time.

src/metadata/extra_tags.rs
```diff
@@ -0,0 +1,365 @@
...
+        // we need to do some magic, because we can have multiple tags pointing to the same arc
+        let mut seen = HashSet::new();
+        for extra_tags in self.0.values() {
+            // only add if this is the first encountered reference to this arc
+            // (using thin pointer equality: https://stackoverflow.com/a/67114787/14681457 ; https://github.com/rust-lang/rust/issues/46139#issuecomment-346971153)
+            if seen.insert(Arc::as_ptr(extra_tags) as *const ()) {
+                if let Err(e) = new_registry.register(extra_tags.clone_arc()) {
+                    panic!("{e}");
+                }
+            }
```
feefladder, line 277: Since we only expect different arcs to come from `register`, which is `Arc::clone`, fat pointers should also compare equal, but there is a bit of a rabbit-hole with there being no guarantee on vtable layouts (IIRC)

tests/sample-exif.tiff
feefladder, line 1: I was a bit unsure to put this here or in `tests/image_tiff/images`, since it's strictly speaking not from there (I _do_ like the example being exif)

PR #120, opened 2025-10-22
Remove broken symlink when building windows wheels

maxrjones:


maxrjones:
wheels were tested in https://github.com/developmentseed/async-tiff/actions/runs/18728327788

kylebarron:
We can't just remove the symlink because that's what's providing the type hints for store classes. The type definition is defined in [this folder in obstore](https://github.com/developmentseed/obstore/tree/main/obstore/python/obstore/_store), and we don't want to duplicate the type definitions here, because they'll get out of sync. 

If we delete the symlink, no Python users on any platform will get any type hinting when they try to create stores via `async_tiff.store`

Ideally the solution would be something like resolving the symlink as part of the windows build process or something like that.

kylebarron:
Or we could even just delete the symlink file as part of the build process on Windows, and then only Windows users won't get type hints, but I don't want to delete the symlink overall.

maxrjones:
does the symlink for type hints only work when installed from the wheels? I tested this out on main where the symlink remains and do not get type hints when using `async_tiff` stores, whereas I do for `obstore` stores.

kylebarron:
Checking it out locally, it looks like that symlink is broken now; I'm not sure why.

From `async-tiff/python/python/async_tiff` you should be able to `ls store` and have it show you the `.pyi` files. 

This worked at one point I'm pretty sure but I'm not sure what's going wrong now and I don't have time to dig into it.

maxrjones:
The symlinks work locally after running `git submodule update --init --recursive`. I updated the PR to only remove the symlink for the windows build. That seems to work -https://github.com/developmentseed/async-tiff/actions/runs/18730296413/job/53425377307

maxrjones:
@kylebarron do you mind if I merge this and trigger a release? This PR unblocks integration with virtualizarr (xref https://github.com/zarr-developers/VirtualiZarr/pull/810)

kylebarron:
Thanks! 

Let me put up another PR quickly that opts in to [abi3](https://www.maturin.rs/bindings#py_limited_apiabi3) wheels, which should make wheel builds faster

PR #124, opened 2025-10-23
chore: Bump python binding version to 0.2.0

kylebarron:


kylebarron:
I'll just get in the pyo3 0.27 upgrade as well, just pending https://github.com/developmentseed/obstore/pull/584

PR #131, opened 2025-11-06
Add support for GeogTOWGS84GeoKey

maxrjones:
This PR fixes the failure related to https://github.com/virtual-zarr/virtual-tiff/issues/52.

I added instructions for downloading the file rather than subsetting and committing the file because gdal drops the problematic tag. 

This looks to be a tag that was proposed as an addition to the GeoTIFF standard but never actually added to the spec - https://github.com/OSGeo/gdal/blob/657c5e4a4fb5b1389fd800ea9cc964435ad89a18/frmts/gtiff/libgeotiff/geokeys.inc#L34.

xref https://trac.osgeo.org/geotiff/wiki/TOWGS84GeoKey

### DIFF
--- a/src/cog.rs
+++ b/src/cog.rs
@@ -23,12 +23,12 @@ mod test {
     use std::io::BufReader;
     use std::sync::Arc;
 
-    use crate::metadata::{PrefetchBuffer, TiffMetadataReader};
-    use crate::reader::{AsyncFileReader, ObjectReader};
-
-    use super::*;
     use object_store::local::LocalFileSystem;
     use tiff::decoder::{DecodingResult, Limits};
+
+    use super::*;
+    use crate::metadata::{PrefetchBuffer, TiffMetadataReader};
+    use crate::reader::{AsyncFileReader, ObjectReader};
 
     #[ignore = "local file"]
     #[tokio::test]
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,44 @@
+name: Lint
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches: ["main"]
+
+permissions: {}
+
+# Make sure CI fails on all warnings, including Clippy lints
+env:
+  CARGO_TERM_COLOR: always
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  clippy_check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Run Clippy
+        run: |
+          cargo clippy --all-targets --all-features
+          cargo clippy --all-targets --all-features --manifest-path python/Cargo.toml
+
+  format_check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Run Rustfmt
+        run: |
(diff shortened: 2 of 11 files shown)


python/src/geo.rs
```diff
@@ -37,7 +37,8 @@ pub(crate) struct PyGeoKeyDirectory {
     geog_azimuth_units: Option<u16>,
     #[pyo3(get)]
     geog_prime_meridian_long: Option<f64>,
-
+    #[pyo3(get)]
```
kylebarron: Would you be able to add a docstring here noting that this isn't part of the spec but is proposed with the link in the PR summary

maxrjones: for sure, added in [`0f4ba6a` (#131)](https://github.com/developmentseed/async-tiff/pull/131/commits/0f4ba6a88867aa4a3990cb873ca65ffe8e9a850b)

src/cog.rs
```diff
@@ -55,6 +55,25 @@ mod test {
         std::fs::write("img.buf", tile).unwrap();
     }
 
+    #[ignore = "local file"]
+    #[tokio::test]
+    async fn tmp_towgs84() {
+        let folder = "/Users/kyle/github/developmentseed/async-tiff";
+        let path = object_store::path::Path::parse("USGS_13_s14w171.tif").unwrap();
```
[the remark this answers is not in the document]
maxrjones: I added a test file in [`e4cecf5` (#131)](https://github.com/developmentseed/async-tiff/pull/131/commits/e4cecf55414746d82e956fdf2fddc4e6ddf19340)

python/src/geo.rs
```diff
@@ -169,6 +173,7 @@ impl From<GeoKeyDirectory> for PyGeoKeyDirectory {
             geog_inv_flattening: value.geog_inv_flattening,
             geog_azimuth_units: value.geog_azimuth_units,
             geog_prime_meridian_long: value.geog_prime_meridian_long,
+            geog_to_wgs84: value.geog_to_wgs84,
```
weiji14:
```suggestion
            geog_to_wgs84_geokey: value.geog_to_wgs84_geokey,
```

tests/geogtowgs_tiff.rs
weiji14: Rename to tests/geogtowgsgeokey_tiff.rs

weiji14:
To be clear, I'm not sure if we should add support for this GeogTOWGS84GeoKey. From reading https://github.com/opengeospatial/geotiff/issues/6, it sounds like 2062 will be reserved, but it doesn't look like it will become part of the GeoTIFF 1.2 spec (whenever that happens). We should however, at least handle the error gracefully instead of panicking.

maxrjones:
> To be clear, I'm not sure if we should add support for this GeogTOWGS84GeoKey. From reading https://github.com/opengeospatial/geotiff/issues/6, it sounds like 2062 will be reserved, but it doesn't look like it will become part of the GeoTIFF 1.2 spec (whenever that happens). We should however, at least handle the error gracefully instead of panicking.

Does this mean that async_tiff would always error on files containing this key, even if the errors are more informative?

weiji14:
> Does this mean that async_tiff would always error on files containing this key, even if the errors are more informative?

Yes, there will be an error. But no, it won't always be an error if we get #100 done, because that means users can pass in a custom tag registry (that could include this GeogTOWGS84GeoKey).

weiji14:
FYI, discussed with Kyle about this offline at FOSS4G. Decided that unknown GeoTags (such as this 2062: GeogTOWGS84GeoKey) can just be skipped, and if someone wants to parse it properly, they can use a custom extension tag registry once that is done for #100.

Issue #41, opened 2025-02-28
Use FIFO instead of LIFO ordering for tile decompression?


kylebarron:
https://docs.rs/tokio-rayon/latest/tokio_rayon/trait.AsyncThreadPool.html#tymethod.spawn_fifo_async

geospatial-jeff:
FIFO is definitely the more common approach and better option here.  Mostly because it provides more consistent / deterministic scheduling.

LIFO is best for specific cases where quickly retrying tasks is important; as any failed tasks are put back on the queue and immediately re-processed.  While an error that is placed back on a FIFO queue must wait in line before being retried.

kylebarron:
Thanks; would you like to put up a PR? should be quick

PR #133, opened 2025-11-22
perf: Change from LIFO to FIFO ordering for tile decoding

weiji14:
In the python `decode_async` method, change from LIFO (Last-In-First-Out) to FIFO (First-In-First-Out) for task execution.

Would be good if there were some benchmarks to compare :slightly_smiling_face: Edit: setting them up at #139.

Fixes https://github.com/developmentseed/async-tiff/issues/41


kylebarron:
This looks great! Feel free to mark as ready for review when ready.

kylebarron:
I think this makes sense, and we should merge this for this release. It would still be great to have benchmarks but we can wait to merge #139 after this release.

Issue #89, opened 2025-04-22
Tokio hangs when opening TIFF


maxrjones:
This example seems to hang, @kylebarron suggested in may relate to the  intersection between exceptions, panics, and async futures

```Python
from async_tiff import TIFF
from async_tiff.store import HTTPStore
from urllib.parse import urlparse
import asyncio

year = 2000
file_name = "50N_120W"
url = f"https://storage.googleapis.com/earthenginepartners-hansen/GLCLU2000-2020/v2/{year}/{file_name}.tif"
parsed = urlparse(url)
store = HTTPStore.from_url(f"{parsed.scheme}://{parsed.netloc}")
async def open_tiff(*, store, path):
    return await TIFF.open(path, store=store)

tiff = asyncio.run(open_tiff(store=store, path=parsed.path))
print(tiff)
```

fyi @TomNicholas

kylebarron:
FWIW I saw this issue https://github.com/apache/arrow-rs-object-store/issues/343 which linked to this upstream issue https://github.com/rust-lang/rust/pull/136932. When I was debugging this issue, I remember seeing a panic somewhere with `u16::Display`. I was super confused at the time, but I'm curious if maybe this was a regression with rustc? Maybe we should test with another rust version

weiji14:
Tested on rust 1.86 and 1.87, getting the same hang behaviour on both so don't think it's related to those issues. I can reproduce this purely on the rust side too (when working on https://github.com/developmentseed/async-tiff/issues/101#issuecomment-2896574826), so at least we can deduce that it's not something with pyo3.

kylebarron:
Yeah... I'm kinda stumped about this hanging 😔

yutannihilation:
After some very primitive print debugging around this function, I found the process doesn't actually hang but it's just slow.

https://github.com/developmentseed/async-tiff/blob/fac684fcd3cb866b64e0f22588283a593433fe57/src/metadata/reader.rs#L257-L280

The fifth tag (and similar tags will come after this) is of 40,000 values.

```
tag_idx: 5
tag_offset: 70, endianness: LittleEndian, bigtiff: false
tag_name: StripOffsets
tag_type_code: 4
tag_type: LONG
count: 40000
value_byte_length: 160000
```

I'm not sure why this is so slow, but it seems each `cursor.read_u32().await?` takes time. Maybe it can fetch values by batch to improve the performance? (I don't know well about the internals. Sorry if it already does!)

[...]

PR #137, opened 2025-11-22
feat: Implement batched tag reading

kylebarron:
Helps but doesn't fully solve https://github.com/developmentseed/async-tiff/issues/89.

plan to more context once tests pass


kylebarron:
Since we added the read-ahead block cache in https://github.com/developmentseed/async-tiff/pull/140, this is no longer necessary. The individual async calls should be nanoseconds of overhead.

PR #139, opened 2025-11-22
test: Setup Rust benchmarks

weiji14:
Benchmark reading a sample Sentinel-2 TCI file, converted to have LZW compression and predictor: 2 (horizontal differencing). Using cargo-codspeed for the benchmarking on GitHub Actions CI.

Adapted from https://github.com/weiji14/foss4g2025/pull/3 and https://github.com/weiji14/foss4g2025/pull/2, but with a smaller GeoTIFF from a different area.

TODO:
- [x] Initial port over
- [x] Split benchmarks to be more unit-like. One to capture network part, another to capture decoding.

References:
- https://codspeed.io/docs/integrations/ci/github-actions

### DIFF
--- a/benches/read_tiff.rs
+++ b/benches/read_tiff.rs
@@ -89,8 +89,15 @@ fn decode_tiff(tiles: Vec<Tile>) -> AsyncTiffResult<Vec<Array>> {
             .map(|tile| tile.decode(&decoder_registry).unwrap())
             .collect()
     });
-    assert_eq!(tile_arrays.len(), 363528192); // 363528192 = 1849 * 196608, should be
-                                              // 361681200 if mask padding is removed
+    assert_eq!(tile_arrays.len(), 1849); // x_count:43 * y_count:43 = 1849
+
+    // Verify byte length by summing all tiled arrays
+    let tile_bytes: usize = tile_arrays
+        .iter()
+        .map(|arr| arr.data().len())
+        .sum::<usize>();
+    assert_eq!(tile_bytes, 363528192); // 363528192 = 1849 * 196608, should be
+                                       // 361681200 if mask padding is removed
 
     Ok(tile_arrays)
 }


weiji14:
Still need to install the GitHub App (https://codspeed.io/docs#connect-your-repository) for the benchmark reports to be uploaded to CodSpeed.

Edit: done.

benches/read_tiff.rs
```diff
@@ -0,0 +1,97 @@
...
+
+            assert_eq!(ifds.len(), 1); // should have only 1 IFD
+            let ifd: &ImageFileDirectory = ifds.first().ok_or(AsyncTiffError::General(
+                "unable to read first IFD".to_string(),
+            ))?;
+
+            let (x_count, y_count) = ifd.tile_count().ok_or(AsyncTiffError::General(
+                "unable to get IFD count".to_string(),
+            ))?;
+            // dbg!(x_count, y_count); // 43 * 43 = 1849
```
kylebarron: Convert into `assert_eq` or remove?

benches/read_tiff.rs
```diff
@@ -0,0 +1,97 @@
...
+        .build()
+        .map_err(|err| AsyncTiffError::External(Box::new(err)))?;
+
+    let tile_bytes: Vec<u8> = pool.install(|| {
+        tiles
+            .into_par_iter()
+            .flat_map_iter(|tile| tile.decode(&decoder_registry).unwrap())
+            .collect()
+    });
+    assert_eq!(tile_bytes.len(), 363528192); // should be 361681200, why not?
```
kylebarron: Hmm what's happening here? Can you explain why you expect the number to be different?

Is it possible there's an issue with whether this is compressed/decompressed? Or whether there's a step in the decoding missing?

weiji14: It should be 10980x10980x3bands = 361681200 bytes (since this is u8 dtype, so 1 byte per pixel). The number is not lower, but higher, so the decompression should have worked, but not sure what's causing the deviation. I should probably try to find a way to piece the bytes together and display it to see what's going on.

kylebarron: Maybe read it in python to inspect it?

kylebarron:
@weiji14 let me know when this is ready for review

benches/read_tiff.rs
```diff
@@ -0,0 +1,97 @@
...
+    let decoder_registry = DecoderRegistry::default();
+
+    // Initialize async runtime
+    let runtime = runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
+
+    // Get list of tiles in TIFF file stream (using tokio async runtime)
+    let tiles: Vec<Tile> = runtime
+        .block_on(async {
```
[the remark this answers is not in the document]
weiji14: Ok, made an `async fn read_tiles()` function at commit 8478a56d62ca86326dd13e56edc2a6ce59520744. But let me do another round of refactoring, because I'd like to separate the async part (network fetch) from the multi-threading part (tile decoding).

weiji14 (commented):
Ready for review @kylebarron! Main change is separate functions for 1) the async network fetching (of compressed bytes) and 2) the multi-threaded decompression/decoding. Lmk if more changes are needed.

kylebarron:
github doesn't understand that "still need to resolve" doesn't mean to resolve

kylebarron:
@weiji14 this should be unblocked because I merged #180

weiji14:
I'm gonna merge this in before you make more breaking changes :laughing:. At least we'll have an end-to-end integration test of some sort now.

kylebarron:
Amazing, thank you!

PR #141, opened 2025-11-22
fix: Handle non utf-8 characters in OME-XML

### LINKED ISSUE #101: OME-TIFF: `Result::unwrap()` on an `Err` value: General("invalid utf-8 sequence of 1 bytes from index 2144")

weiji14:
By using [`String::from_utf8_lossy`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy).

Changes:
- Used another sample OME-TIFF file that has non utf8 characters in the OME-XML
- ~Increased prefetch bytes to more than 32kb as workaround to avoid hanging when parsing the StripOffsets tag (xref #89)~ Edit: Not needed anymore after https://github.com/developmentseed/async-tiff/pull/140.

Fixes #101


kylebarron:
Ok so to clarify the spec says that these characters _should_ always be UTF8? And which spec are we talking about? TIFF or OME-TIFF?

This tag type is named `Value:Ascii`, so it seems like a mismatch that we're hitting utf8 issues. Should we rename that to `Value::Text`? or `Value::String`?

Alternatively we could store a `Vec<u8>` but if this is string data, then `Vec<u8>` seems improper

weiji14:
> Ok so to clarify the spec says that these characters _should_ always be UTF8? And which spec are we talking about? TIFF or OME-TIFF?

OME-TIFF spec says so, see https://docs.openmicroscopy.org/ome-model/5.6.3/ome-tiff/specification.html

> This tag type is named `Value:Ascii`, so it seems like a mismatch that we're hitting utf8 issues. Should we rename that to `Value::Text`? or `Value::String`?

I think this type is something in the TIFF spec, but not 100% sure. Probably ok to just keep it as ASCII.

> Alternatively we could store a `Vec<u8>` but if this is string data, then `Vec<u8>` seems improper

Yeah, I reckon using a String is better here, otherwise users will need to decode bytes to string...

PR #143, opened 2025-11-23
refactor(ExponentialMetadataCache): rename & make configurable

alukach:
rename ExponentialMetadataCache to ReadAheadMetadataCache and update to support configuration


kylebarron:
Thanks for the PR! I incorporated these ideas into #140

Issue #173, opened 2026-01-23
Fix decoding of edge tiles


kylebarron:
Erm, I think you broke decoding of edge tiles in #165. I'm getting errors like this:

```
thread '<unnamed>' (80345) panicked at benches/read_tiff.rs:89:56:
called `Result::unwrap()` on an `Err` value: General("Internal error: incorrect shape or data length passed to Array::try_new. Got data length 196608, expected 175104")
```

The decoding (which includes decompression, reversing the predictors, etc) needs to operate on the unclipped tile_width and tile_height, rather than the clipped width/height given by [`compute_tile_dimensions`](https://github.com/developmentseed/async-tiff/blob/933c910738ec0f18a95ad688367afa4957b55e5b/src/ifd.rs#L831-L857)

_Originally posted by @weiji14 in https://github.com/developmentseed/async-tiff/pull/139#discussion_r2719341474_

PR #180, opened 2026-01-23
fix: Add handling of unaligned tiles on image border

kylebarron:
### Change list

- Fix how we read and decode the unaligned image tiles in the pixel remainder
- Add test for reading such a TIFF
- This only edits unrelated tests so that it can change the signature of `open_tiff` in the test utils.

Closes https://github.com/developmentseed/async-tiff/issues/173

Should unblock https://github.com/developmentseed/async-tiff/pull/139


kylebarron:
The only unrelated changes are just modifying the test paths now that we need to fetch fixtures from our geotiff test data submodule too

Issue #154, opened 2025-12-09
Support LERC decompression


kylebarron:
There's an Apache-2 licensed Rust library (https://docs.rs/lerc/latest/lerc/) that wraps the LERC C++ library.

kylebarron:
Looks like this will take a little bit of work because the existing decoder API only works in terms of `u8` while LERC is more strongly typed. First you can use [`get_blob_info`](https://docs.rs/lerc/latest/lerc/fn.get_blob_info.html) to infer the [`BlobInfo`](https://docs.rs/lerc/latest/lerc/struct.BlobInfo.html) including the data type and array size. 

Then you need to call [`decode_with_info`](https://docs.rs/lerc/latest/lerc/fn.decode_with_info.html) _already knowing_ the `T: LercDataType`.

So we may want to bring the typed data higher up into the Decoder. The Decoder shouldn't _have_ to decode into a strongly-typed buffer, but it should _be able to_ if it so chooses.

kylebarron:
Ref https://github.com/geotiffjs/geotiff.js/pull/206

Issue #203, opened 2026-01-29
Refactor Decoder trait


kylebarron:
- Pass in a dataclass of parameters/attributes instead of `PhotometricInterpretation` and JPEG tables as top-level args
- Pass in output data type?
- Allow returning more strongly-typed data rather than just `Vec<u8>`, ref https://github.com/developmentseed/async-tiff/issues/154

feefladder:
One comment on:

>     * Allow returning more strongly-typed data rather than just `Vec<u8>`, ref [Support LERC decompression #154](https://github.com/developmentseed/async-tiff/issues/154)

I fully support more strongly-typed arrays from decoder. I had a small nitpick in #234, that would be solved by that.

When thinking about a `Decoder` trait with two possible implementations (typed and not-typed), I couldn't think of a way to implement two different functions bidirectionally. Exposing a `fn(buf: &mut [u8])` to Python doesn't work, so I think the trait can be split in two with a blanket implementation:
```rust
pub struct DecodingInfo {
  photometric_interpretation: PhotometricInterpretation,
  jpeg_tables: Option<&[u8]>,
  samples_per_pixel: u16,
  bits_per_sample: u16, // this is still needed in case data_type is unknown
  data_type: Option<DataType>,
  n_pixels: usize // the number of pixels in this tile, including padding
}

/// A trait to decode a TIFF tile. This is exposed to Python
pub trait TypedDecoder: Debug + Send + Sync {
    /// Decode a TIFF tile.
    fn decode_tile_typed(
        &self,
        buffer: Bytes,
        decoding_info: DecodingInfo
    ) -> AsyncTiffResult<TypedArray>;
}

/// The trait that all current implementations can easily be modified to implement
pub trait Decoder: Debug + Send + Sync {
    fn decode_tile(
        &self,
        buffer: Bytes,
        out_buffer: &mut[u8],
        decoding_info: DecodingInfo
    ) -> AsyncTiffResult<()>;
}

/// Blanket implementation RawDecoder=>Decoder
impl<T: Decoder> TypedDecoder for T {
    fn decode_tile_typed(
        &self,
        buffer: Bytes,
        decoding_info: DecodingInfo
    ) -> AsyncTiffResult<TypedArray>  {
        let typed_array = TypedArray::new(&decoding_info);
        self.decode_tile_raw(buffer, typed_array.as_mut(), decoding_info)
        typed_array
    }
}
```
[...]

PR #204, opened 2026-01-29
feat: support LERC, LERC_DEFLATE, LERC_ZSTD decompression

kylebarron:
Closes https://github.com/developmentseed/async-tiff/issues/154. 

LERC is only able to load as _typed arrays_. So this will benefit from the refactor in https://github.com/developmentseed/async-tiff/issues/203. For now this incurs extra casts/copies to go from typed array to uint8 and then back to a typed array.


kylebarron:
Thank you @zakjan for your hint in https://github.com/geotiffjs/geotiff.js/pull/206!

PR #241, opened 2026-02-05
feat:metadata expose `MetadataCursor` and add accessors for `ImageFileDirectoryReader`

feefladder:
#82 gives the user control over metadata reading, and some methods of `MetadataCursor` were public, but not the type itself. This allows the user to read tag type, count and value by themselves if they want to do complex metadata caching.

### DIFF
--- a/src/metadata/reader.rs
+++ b/src/metadata/reader.rs
@@ -215,8 +215,7 @@ impl ImageFileDirectoryReader {
         tag_idx: u64,
     ) -> AsyncTiffResult<(Tag, TagValue)> {
         assert!(tag_idx < self.tag_count);
-        let tag_offset =
-            self.ifd_start_offset + self.tag_count_byte_size + (self.ifd_entry_byte_size * tag_idx);
+        let tag_offset = self.tag_offset(tag_idx);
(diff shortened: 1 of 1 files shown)


kylebarron:
Thanks; I need to think about this though. There's a proliferation of public APIs and I'd much rather figure out how to remove public APIs than add new ones.

feefladder:
I understand. there's two pathways I see here:
1. minimize methods on MetadataCursor (`new()`,`with_offset`,`new_with_offset`,`seek`)
2. Don't expose MetadataCursor at all.
   - add a `ImageFileDirectoryReader.peek_tag(&self, tag_offset: u64) -> (Tag, TagType, count: u64, Option<offset: u64>)` function.
   - This could stand on its own, or (Optional) refactor the `read_tag`,`read_tag_value` functions:
     1. peek_tag, if Some(offset) -> move cursor
     2. read tag value from cursor
     But I think the refactor loses the scope of this PR and would fit a separate PR (if at all).
   - I've made a sample implementation [here](https://github.com/feefladder/async-tiff/blob/peek-tag/src/metadata/reader.rs#L230) for the `peek_tag` function.

PR #264, opened 2026-03-04
Use uncropped tile width for predictor

kylebarron:
## What I am changing

-

## How I did it

-

## How you can test it

-

## Related Issues

-


kylebarron:
I think this is superseded by #265

PR #265, opened 2026-03-04
refactor: refactor predictor.rs for simplicity

kylebarron:
- Remove `PredictorInfo` struct. That carries a lot more data than we actually need for the predictor handling.
- Removes the predictor tests, which I really don't understand anyways. We should be testing against real data for accuracy instead of contrived examples.
- Add integration tests with rasterio tile reading

### DIFF
--- a/python/tests/test_integration_rasterio.py
+++ b/python/tests/test_integration_rasterio.py
@@ -41,9 +41,11 @@ if TYPE_CHECKING:
         ("rasterio", "uint8_rgb_deflate_block64_cog"),
         ("rasterio", "uint8_rgb_webp_block64_cog"),
         ("rasterio", "uint8_rgba_webp_block64_cog"),
-        ("rio-tiler", "cog_rgb_with_stats"),
+        # Ycbcr subsampling not implemented
+        # ("rio-tiler", "cog_rgb_with_stats"),
         ("umbra", "sydney_airport_GEC"),
-        ("vantor", "maxar_opendata_yellowstone_visual"),
+        # Ycbcr subsampling not implemented
+        # ("vantor", "maxar_opendata_yellowstone_visual"),
     ],
 )
 async def test_read(


mdsumner:
I just stomped on this (I believe). the attached tiff file has four tiles (ifd level 0 only) - one with deflate predictor  = 2, one without predictor -  and each 16x16 tile is filled with pixel index (but right most colum and bottom row are cropped). This messes up the decoding, as also seen with 

```
https://e84-earth-search-sentinel-data.s3.us-west-2.amazonaws.com/sentinel-2-c1-l2a/55/G/DN/2026/2/S2C_T55GDN_20260227T000650_L2A/B04.tif
```

tiff creation in R

```R
## blocks must be 16x16 at minimum so
dm <- c(16, 16)
v <- seq_len(prod(dm))
m <- matrix(v, dm[2], dm[1], byrow = TRUE)

nc <- cbind(m, m)
library(terra)
r <- rast(rbind(nc, nc)[seq(1, 2 * dm[1] - 1), seq(1, 2 * dm[2] - 1)])
plot(r)
filename <- "deflate-16x16-crop-no-predictor.tif"

writeRaster(r, filename, 
            gdal = c("BLOCKXSIZE=16", "BLOCKYSIZE=16", "TILED=YES"), 
            filetype = "GTiff", overwrite = TRUE)

filename2 <- "deflate-16x16-crop-predictor=2.tif"
writeRaster(r, filename2, 
            gdal = c("BLOCKXSIZE=16", "BLOCKYSIZE=16", "TILED=YES", "COMPRESS=DEFLATE", "PREDICTOR=2"), 
            filetype = "GTiff", overwrite = TRUE)

```

 (I could craft non-R version of this if that's useful), or any other testing I'd be happy to see this fixed - and confirmed my stance on what the problem is! 

decode-with-predictor
decode-without-predictor

[deflate-16x16-tiff.zip](https://github.com/user-attachments/files/25921455/deflate-16x16-tiff.zip)

sentinel-2-br-ifd0_S2C_T55GDN_20260227T000650_L2A

mdsumner:
just to confirm - it's fixed, thanks!     from ebb6664ba5a0c9640760b1e3c71ec668d7f82460

image

kylebarron:
Just to confirm, @mdsumner do you have an issue with the latest release or not?

mdsumner:
No, it's fixed in latest. I was in the middle of investigating and found this PR. Pulling latest and confirming was easier than I expected! Maybe the test tifs are useful?

kylebarron:
If you find any _errors_, then creating a new issue and uploading small, reproducibly-created example TIFFs to https://github.com/developmentseed/geotiff-test-data would be helpful.

PR #268, opened 2026-03-04
ci: Cancel in-progress workflows when new commits are pushed

kylebarron:
I learned that async-tiff is our most expensive repo for Github Actions currently (I didn't know we paid for GHA on open repos), so we should at least implement concurrency limits.


weiji14:
Huh, I thought it was free too. Can we check the billing report to see what's going on? Let's chat internally.

PR #269, opened 2026-03-04
chore: add plausible analytics for async-tiff

aboydnw:
Configure MkDocs analytics to use Plausible with the async-tiff site identifier.

Similar to what we've done on titiler and lonboard


kylebarron:
I _think_ you also need this file... https://github.com/developmentseed/lonboard/blob/main/docs/overrides/partials/integrations/analytics/plausible.html

Or maybe that's only for adding a 👍 👎 was this helpful at the bottom of the page?

aboydnw:
Ah yeah good catch @kylebarron , I think I added now

PR #311, opened 2026-06-23
feat: Relax 'Send' on `[async-trait]`

dmarteau:
## Description

Atm, async-tiff fail to compile for `wasm32-unknown-unknown` target  because of the `Send` requirements on `Futures` implied by the `[async_trait]` decoration which is not set for  `wasm_bindgen_futures::JsFuture`.

That is, when using   a reader that return `wasm_bindgen_futures::JsFuture` (like `reqwest` does when compiling for `wasm32-unknown-unknown` the compilation fail with the following error:

```
error[E0277]: `Rc<RefCell<js_sys::futures::Inner>>` cannot be sent between threads safely
      --> /home/david/.cargo/git/checkouts/async-tiff-c522a11303546031/ce7753d/src/reader.rs:216:5
        |
216 |     async fn get_bytes(&self, range: Range<u64>) -> AsyncTiffResult<Bytes> {
        |     ^^^^^
        |
        |     `Rc<RefCell<js_sys::futures::Inner>>` cannot be sent between threads safely
        |     within this `{async block@/home/david/.cargo/git/checkouts/async-tiff-c522a11303546031/ce7753d/src/reader.rs:216:5: 216:10}`
        |
        = help: within `{async block@/home/david/.cargo/git/checkouts/async-tiff-c522a11303546031/ce7753d/src/reader.rs:216:5: 216:10}`, the trait `std::marker::Send` is not implemented for `Rc<RefCell<js_sys::futures::Inner>>`
note: required because it appears within the type `js_sys::futures::JsFuture`
      --> /home/david/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/mod.rs:118:12
        |
118 | pub struct  JsFuture<T = JsValue> {
        |                     ^^^^^^^^
note: required because it's used within this `async` fn body  
      --> /home/david/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/reqwest-0.13.4/src/wasm/mod.rs:34:1

...
```

## Solution
This can be fixed by relaxing the `Send`  constraint set by `[async_trait]` by using the `[async_trait(?Send)]`  annotation.

However I don't know if this appropriate in all cases when `Send` may be required on `Futures`. 
In this case, I wonder how te use a target configuration option without duplicating code (i.e that applies only on the `[async-trait]` decoration.

Thx for the job done with this crate !

### DIFF
--- a/.github/workflows/wasm-build.yml
+++ b/.github/workflows/wasm-build.yml
@@ -1,0 +1,30 @@
+name: Wasm
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  wasm_build:
+    name: Wasm build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: "recursive"
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - run: cargo install wasm-pack
+
+      - name: Wasm target compilation
+        run: wasm-pack build
--- a/.github/workflows/wasm-build.yml
+++ b/.github/workflows/wasm-build.yml
@@ -24,7 +24,5 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - run: cargo install wasm-pack
-
       - name: Wasm target compilation
-        run: wasm-pack build
+        run: cargo check --target wasm32-unknown-unknown --lib
--- a/.github/workflows/wasm-build.yml
+++ b/.github/workflows/wasm-build.yml
@@ -6,6 +6,8 @@ on:
       - main
   pull_request:
 
+permissions: {}
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
@@ -17,10 +19,12 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          submodules: "recursive"
+          persist-credentials: false
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
 
       - uses: Swatinem/rust-cache@v2
 
--- a/src/metadata/cache.rs
+++ b/src/metadata/cache.rs
@@ -143,7 +143,8 @@ impl<F: MetadataFetch> ReadaheadMetadataCache<F> {
     }
 }
 
-#[async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 impl<F: MetadataFetch + Send + Sync> MetadataFetch for ReadaheadMetadataCache<F> {
     async fn fetch(&self, range: Range<u64>) -> AsyncTiffResult<Bytes> {
         let mut cache = self.cache.lock().await;
@@ -190,7 +191,8 @@ mod test {
         }
     }
 
-    #[async_trait]
+    #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+    #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
     impl MetadataFetch for TestFetch {
         async fn fetch(&self, range: Range<u64>) -> crate::error::AsyncTiffResult<Bytes> {
             if range.start as usize >= self.data.len() {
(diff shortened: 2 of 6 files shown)


kylebarron:
I don't think removing `Send` on async-trait is realistic. The Python bindings require `Send`.

In theory you could implement something like [`MaybeSend`](https://github.com/zarrs/zarrs/blob/38a7be3e51c0b7f2f6a88ba0859714ab07878cb4/zarrs_plugin/src/maybe.rs#L5-L28) that requires `Send` except for on `wasm32-unknown-unknown`, but not sure how that relates to async-trait.

But stepping back a sec, why do you want to compile `async-tiff` to `wasm32-unknown-unknown`? There are excellent pure TypeScript TIFF/GeoTIFF async readers in TypeScript, including (my own) [`@developmentseed/geotiff`](https://github.com/developmentseed/deck.gl-raster/tree/main/packages/geotiff). Or if you don't want `GeoTIFF` semantics, you could use [`@cogeotiff/core`](https://github.com/blacha/cogeotiff), which supports generic `TIFF` files.

dmarteau:
> I don't think removing `Send` on async-trait is realistic. The Python bindings require `Send`.

I gated the option behind a `cfg_attr`, so that `Send` is only removed when the feature `nosendfutures` is set - which fix the Python  bindings.

> But stepping back a sec, why do you want to compile `async-tiff` to `wasm32-unknown-unknown`? 

We are working on a wgpu-based asynchronous map renderer that should run natively  and on the web, so it must be pure Rust - Geotiff support  is only a part of it.

kylebarron:
I don't think I want to add a new feature for this. Could you change it to be platform based, explicitly check for wasm32 [as zarrs does it](https://github.com/zarrs/zarrs/blob/38a7be3e51c0b7f2f6a88ba0859714ab07878cb4/zarrs_plugin/src/maybe.rs#L5-L28)

dmarteau:
> I don't think I want to add a new feature for this. Could you change it to be platform based, explicitly check for wasm32 [as zarrs does it](https://github.com/zarrs/zarrs/blob/38a7be3e51c0b7f2f6a88ba0859714ab07878cb4/zarrs_plugin/src/maybe.rs#L5-L28)

If you do that, you duplicate all code using `[async_trait]` annotation:  the constraint lies in the `[async_trait]` generated code  not in the annotated code itself - and this is why the `(?Send)`   has been added.   I don't think you want that either.

src/metadata/cache.rs
```diff
@@ -143,7 +143,8 @@ impl<F: MetadataFetch> ReadaheadMetadataCache<F> {
     }
 }
 
-#[async_trait]
+#[cfg_attr(feature = "nosendfutures", async_trait(?Send))]
+#[cfg_attr(not(feature = "nosendfutures"), async_trait)]
```
kylebarron: Right now you have cfg_attr based on a feature, isn't there a way to do this `cfg_attr` based on an architecture?

dmarteau: Apparently you can: https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute

dmarteau: Changed ` #[cfg_attr(feature = "nosendfutures", ... `  to `#[cfg_attr(target_arch = "wasm32". ...`  . Compilation to wasm32 is ok.

kylebarron:
Thanks. Would you be able to add a test setup or a CI step that checks compilation in `wasm32-unknown-unknown`? Otherwise I won't notice if something breaks that compilation in the future.

dmarteau:
> Thanks. Would you be able to add a test setup or a CI step that checks compilation in `wasm32-unknown-unknown`? Otherwise I won't notice if something breaks that compilation in the future.

I'm on it but my bandwidth will be quite narrow for the next few days.

dmarteau:
> Thanks. Would you be able to add a test setup or a CI step that checks compilation in `wasm32-unknown-unknown`? Otherwise I won't notice if something breaks that compilation in the future.

Added a github workflow for compiling to WASM target

.github/workflows/wasm-build.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - run: cargo install wasm-pack
+
+      - name: Wasm target compilation
+        run: wasm-pack build
```
kylebarron: Do we need to be using `wasm-pack build`? That's a lot of overhead; can't we just use `cargo build --target wasm32-unknown-unknown`?

kylebarron: I took out the `wasm-pack build` in https://github.com/developmentseed/async-tiff/pull/311/changes/3757e54076c209dbabbc6e3e73439bbf26b187b5, which also removed the `Cargo.toml` change

kylebarron:
Thanks!

PR #340, opened 2026-09-11
feat: support sparse GeoTIFF tiles

ljstrnadiii:
## Summary

GDAL/rasterio support "sparse" tiled TIFFs (`SPARSE_OK=YES`): a tile the encoder never wrote has `TileOffsets[i] == TileByteCounts[i] == 0`. Fetching one currently errors instead of decoding to a nodata-filled array:

```
TypeError: Generic LocalFileSystem error: Requested range was invalid
```

`TileByteRange::from_ifd_tile` builds `offset..offset+byte_count` unconditionally, so a sparse tile becomes the range `0..0`, which `object_store` rejects outright. No decoder can turn 0 bytes into a tile either.

## Changes

- `TileByteRange`/`TilesByteRanges::into_fetch` treat any empty range as sparse rather than fetching it, per band — a planar tile with some bands populated and others sparse works, not treated as malformed.
- `CompressedBytes::Chunky`/`Planar` now hold `Option<Bytes>` (`None` == sparse).
- `Tile` gained `nodata: Option<f64>`, parsed from the IFD's `GDAL_NODATA` tag at construction.
- `Tile::decode` fills sparse tiles/bands with that value (or `0`) instead of touching the decoder, per band.
- Predictor is now applied per-band for planar tiles (`samples = 1`), not once across the whole buffer with `samples = total-band-count` — the latter only happened to be correct for single-band planar tiles.
- New `Tile::is_sparse()` / `ImageFileDirectory::is_tile_sparse()`, exposed to Python too.

Commits: fail → fix → python bindings.

## Status

Ready for review. CI green. Blocked from merging by developmentseed/geotiff-test-data#70 — the submodule is pinned to a commit that only exists on that PR's branch, which will get a new SHA once it merges (likely squashed).

### Pre-merge checklist
- [ ] Bump the `fixtures/geotiff-test-data` submodule pin to developmentseed/geotiff-test-data#70's post-merge commit.
- [ ] Don't merge before developmentseed/geotiff-test-data#70.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/python/async_tiff/_ifd.pyi
+++ b/python/python/async_tiff/_ifd.pyi
@@ -172,4 +172,15 @@ class ImageFileDirectory:
 
         This will be `None` if the IFD is not tiled.
         """
+    def is_tile_sparse(self, x: int, y: int) -> bool:
+        """Whether the tile at the given column and row is entirely sparse -- i.e. never
+        written by the encoder. `False` for a planar tile with only some bands sparse.
+
+        Args:
+            x: The column index within the ifd to check.
+            y: The row index within the ifd to check.
+
+        Raises:
+            ValueError: if the IFD is not tiled.
+        """
         ...
--- a/python/python/async_tiff/_tile.pyi
+++ b/python/python/async_tiff/_tile.pyi
@@ -14,15 +14,24 @@ class Tile:
     def y(self) -> int:
         """The row index this tile represents."""
     @property
-    def compressed_bytes(self) -> Buffer | list[Buffer]:
+    def compressed_bytes(self) -> Buffer | None | list[Buffer | None]:
         """The compressed bytes underlying this tile.
 
         This will be a single buffer for pixel-interleaved (chunky) data, or a list of
-        buffers for band-interleaved (planar) data.
+        buffers for band-interleaved (planar) data. A `None` entry means that tile or band
+        was never written by the encoder (sparse); see `is_sparse`.
         """
     @property
     def compression_method(self) -> Compression | int:
         """The compression method used by this tile."""
+    @property
+    def is_sparse(self) -> bool:
+        """Whether this tile is entirely sparse (never written by the encoder).
+
+        `False` for a planar tile with only some bands sparse; inspect
+        `compressed_bytes` for per-band detail in that case. `decode` still succeeds for a
+        sparse tile, filling it with the IFD's nodata value (or `0`).
+        """
     def decode_sync(
         self,
         *,
--- a/python/src/tile.rs
+++ b/python/src/tile.rs
@@ -56,6 +56,14 @@ impl PyTile {
             .map(|t| t.compression_method().into())
     }
 
+    #[getter]
+    fn is_sparse(&self) -> PyResult<bool> {
+        self.0
+            .as_ref()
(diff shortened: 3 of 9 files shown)


kylebarron:
Can you update the geotiff-test-data git submodule to reflect the squashed commit?

python/python/async_tiff/_ifd.pyi
```diff
@@ -172,4 +172,15 @@ class ImageFileDirectory:
 
         This will be `None` if the IFD is not tiled.
         """
+    def is_tile_sparse(self, x: int, y: int) -> bool:
+        """Whether the tile at the given column and row is entirely sparse -- i.e. never
+        written by the encoder. `False` for a planar tile with only some bands sparse.
```
kylebarron, line 177: This should link to the GDAL docs about sparse tiles, as well as explicitly defining here that a sparse tile has a byte length of 0.

It's also true that a sparse tile is more poorly defined for planar tiffs

python/python/async_tiff/_tile.pyi
```diff
@@ -14,15 +14,24 @@ class Tile:
...
     @property
     def compression_method(self) -> Compression | int:
         """The compression method used by this tile."""
+    @property
+    def is_sparse(self) -> bool:
+        """Whether this tile is entirely sparse (never written by the encoder).
+
+        `False` for a planar tile with only some bands sparse; inspect
+        `compressed_bytes` for per-band detail in that case. `decode` still succeeds for a
+        sparse tile, filling it with the IFD's nodata value (or `0`).
```
kylebarron, line 33: I'd prefer for public-facing docstrings not to be AI-written

python/python/async_tiff/_tile.pyi
```diff
@@ -14,15 +14,24 @@ class Tile:
...
         This will be a single buffer for pixel-interleaved (chunky) data, or a list of
-        buffers for band-interleaved (planar) data.
+        buffers for band-interleaved (planar) data. A `None` entry means that tile or band
+        was never written by the encoder (sparse); see `is_sparse`.
         """
     @property
     def compression_method(self) -> Compression | int:
         """The compression method used by this tile."""
+    @property
+    def is_sparse(self) -> bool:
```
kylebarron, line 28: I'm not sure whether async-tiff should necessarily adopt the _GDAL-specific_ interpretation that a tile with no byte length is sparse. We should consult the TIFF spec to see what it allows. 

Perhaps even sparse tiles is something that should only exist as a concept in async-geotiff not async-tiff. I'm not sure,

python/src/tile.rs
```diff
@@ -102,17 +110,22 @@ impl PyTile {
 
 #[derive(IntoPyObject)]
 enum PyCompressedBytes {
-    Chunky(PyBytes),
-    Planar(Vec<PyBytes>),
+    /// `None` if the tile is sparse (never written by the encoder).
+    Chunky(Option<PyBytes>),
+    /// `None` per band that's sparse.
+    Planar(Vec<Option<PyBytes>>),
```
kylebarron, line 116: I don't want to do this. The _bytes_ should never be nullable. Whatever is creating these bytes should have nullability encoded there

PR #341, opened 2026-09-18
chore: Prepare rust 0.4 release

weiji14:
Bump version and include changelog (have filtered out all the `chore (deps)` entries to keep things focused).

Previous release at #293.


kylebarron (approved):
Thanks!

kylebarron:
I think you should be able to just tag and push to main and it should deploy a new version.

